PR #6, opened 2018-09-21
Fix data dir configuration for simple cluster

scholzj:
The simple cluster setup in Getting started is missing configuration for the directories where data will be stored (which are created during installation). This PR adds this information. It also fixes wrong config file name used when starting Kafka.

### DIFF
--- a/books/proc-running-single-node-amq-streams-cluster.adoc
+++ b/books/proc-running-single-node-amq-streams-cluster.adoc
@@ -53,7 +53,7 @@ log.dirs=/var/lib/kafka/
 [source,shell,subs=+quotes]
 ----
 su - kafka
-/opt/kafka/bin/kafka-server-start.sh -daemon /opt/kafka/config/kafka.properties
+/opt/kafka/bin/kafka-server-start.sh -daemon /opt/kafka/config/server.properties
 ----
 
 . Make sure that Apache Kafka is running.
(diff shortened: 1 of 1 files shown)


books/proc-running-single-node-amq-streams-cluster.adoc
```diff
@@ -17,7 +17,15 @@ WARNING: Single node AMQ Streams cluster does not provide realibility and high a
 
 .Running the cluster
 
-. Start Apache Zookeeper with the default configuration file.
+. Edit the Zookeeper configuration file `/opt/kafka/config/zookeeper.properties`.
+Set the option `dataDir` to `/var/lib/zookeeper/`.
```
dmuntima: Set the option `dataDir` to > Set the `dataDir` option to

books/proc-running-single-node-amq-streams-cluster.adoc
```diff
@@ -32,12 +40,20 @@ su - kafka
 jcmd | grep zookeeper
 ----
 
-. Start Apache Kafka with the default configuration file.
+. Edit the Kafka configuration file `/opt/kafka/config/server.properties`.
+Set the option `log.dirs` to `/var/lib/kafka/`.
```
dmuntima: Set the option `log.dirs` to  > Set the `log.dirs` option to

books/proc-running-single-node-amq-streams-cluster.adoc
```diff
@@ -32,12 +40,20 @@ su - kafka
...
++
+[source,properties,subs=+quotes]
+----
+log.dirs=/var/lib/kafka/
+----
+
+. Start Apache Kafka.
```
dmuntima: Start Apache Kafka. >  Start  Kafka. (I think we decided not use Apache whenever possible.)

PR #7, opened 2018-09-21
Add Zookeeper configuration section

scholzj:
This PR adds the Zookeeper configuration section.

### DIFF
--- a/books/con-zookeeeper-cluster-configuration.adoc
+++ b/books/con-zookeeeper-cluster-configuration.adoc
@@ -6,24 +6,24 @@
 
 = Zookeeper cluster configuration
 
-For production use cases it is strongly recommended to run a cluster of replicated Zookeeper instances.
-
-TIP: Zookeeper clusters are sometimes also refer to as _ensembles_.
+For reliable ZooKeeper service, you should deploy ZooKeeper in a cluster.
+Hence, for production use cases, you must run a cluster of replicated Zookeeper instances.
+Zookeeper clusters are also referred as ensembles.
 
 Zookeeper clusters usually consist of an odd number of nodes.
-Zookeeper requires a majority of cluster nodes to be available in order to work. 
-For example, a cluster with 3 nodes requires at least 2 of them to be up and running. 
-In other words, it can tolerate 1 node being down. 
-A cluster consisting of 5 nodes requires at least 3 nodes to be available. 
-In other words, it can tolerate 2 nodes being down.
-A cluster consisting of 7 nodes requires at least 4 nodes to be available. 
-In other words, it can tolerate 3 nodes being down.
+Zookeeper requires a majority of the nodes in the cluster should be up and running.
+For example, a cluster with three nodes, at least two of them must be up and running.
+This means it can tolerate one node being down.
+A cluster consisting of five nodes, at least three nodes must be available.
+This means it can tolerate two nodes being down.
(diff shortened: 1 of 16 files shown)


books/con-zookeeeper-cluster-configuration.adoc
```diff
@@ -0,0 +1,72 @@
...
+The time is specified as a number of ticks (see the `timeTick` option for more details).
+`syncLimit`:: Amount of time for which followers can be behind the leader.
+The time is specified as a number of ticks (see the `timeTick` option for more details).
+
+In addition to the options above, every configuration file should contain a list of servers which should be members of the Zookeeper cluster. 
+The server records should be specified in the format `server.id=hostname:port1:port2`, where:
+
+`id`:: is the ID of the Zookeeper cluster node.
+`hostname`:: is the hostname or IP address where the node listens for connections.
+`port1`:: is the number of the port used for intra-cluster communication.
```
ppatierno: often referred as "peer" port? do we want to say "intra-cluster/peer"?

books/con-zookeeeper-cluster-configuration.adoc
```diff
@@ -0,0 +1,72 @@
...
+clientPort=2181
+initLimit=5
+syncLimit=2
+
+server.1=172.17.0.1:2888:3888
+server.2=172.17.0.2:2888:3888
+server.3=172.17.0.3:2888:3888
+----
+
+Each node in the Zookeeper cluster has to be assigned an `ID`.
```
ppatierno: should we start to talk about Zookeeper ID before the config file where IDs are used for listing the nodes in the cluster?

scholzj: If you tell me how to move it, I can move it. But I think it will always talk about something what is _not defined_ yet. If I start with ID on the beginning, I will either talk about the `server.` options which I did not explain or I will be unable to explain what is the purpose of this. So IMHO it is ok as it is.

books/con-zookeeeper-cluster-configuration.adoc
scholzj: I never heard it being refereed to as _peer port_. _Peer_ port is at least for me completely confusing. Since I personally would not understand _peer communication_ as a necessarily intra-cluster communications. So I would prefer to keep it simple and as it is.

books/con-zookeeper-basic-configuration.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+= Basic configuration
+
+The most important Zookeepr configuration options are:
+
+`tickTime`:: Zookeeper's basic time unit (in milliseconds) used for heartbeats and session timeouts (for example, minimum session timeout will be two ticks).
+`dataDir`:: The directory where Zookeeper stores its transaction log and snapshots of its in-memory database. This should be set to the `/var/lib/zookeeper/` directory created during installation.
+`clientPort`:: Port number where clients can connect. Defaults to `2181`.
+
+NOTE: A sample configuration file can be found in `config/zookeeper.properties` in the AMQ Streams installation directory.
```
dmuntima: Delete the admonitions and move the text as a paragraph. Rewrite to: 
An example zookeeper configuration file `config/zookeeper.properties `is located in the AMQ Streams installation directory.  It is recommended to place the `dataDir` directory on a separate disk device to minimize the latency in Zookeeper.

PR #9, opened 2018-09-24
Add chapter about topics

scholzj:
This PR adds a section about topics and topic management.

### DIFF
--- a/books/con-internal-topics.adoc
+++ b/books/con-internal-topics.adoc
@@ -1,0 +1,33 @@
+// Module included in the following assemblies:
+//
+// assembly-topics.adoc
+
+[id='con-internal-topics-{context}']
+
+= Internal topics
+
+Kafka has several internal topics.
+These are used to store consumer offsets (`\__consumer_offsets`) or transaction state (`__transaction_state`).
+These topics can be configured using dedicated options starting with prefix `offsets.topic.` and `transaction.state.log.`. 
+The most important configuration options are:
+
+`offsets.topic.replication.factor`::
+Number of replicas for `__consumer_offsets` topic.
+The default value is `3`.
+
+`offsets.topic.num.partitions`::
+Number of partitions for `__consumer_offsets` topic.
+The default value is `50`.
+
+`transaction.state.log.replication.factor`::
+Number of replicas for `__transaction_state` topic.
+The default value is `3`.
+
+`transaction.state.log.num.partitions`::
+Number of partitions for `__transaction_state` topic.
+The default value is `50`.
+
+`transaction.state.log.min.isr`::
+Minimum number of replicas that must acknowledge a write to `__transaction_state` topic to be considered successful.
+If this minimum cannot be met, then the producer will fail with an exception.
+The default value is `2`.
--- a/books/con-internal-topics.adoc
+++ b/books/con-internal-topics.adoc
@@ -8,7 +8,7 @@
 
 Kafka has several internal topics.
 These are used to store consumer offsets (`\__consumer_offsets`) or transaction state (`__transaction_state`).
-These topics can be configured using dedicated options starting with prefix `offsets.topic.` and `transaction.state.log.`. 
+These topics can be configured using dedicated Kafka broker configuration options starting with prefix `offsets.topic.` and `transaction.state.log.`. 
 The most important configuration options are:
 
 `offsets.topic.replication.factor`::
--- a/books/con-internal-topics.adoc
+++ b/books/con-internal-topics.adoc
@@ -6,6 +6,7 @@
 
 = Internal topics
 
+Internal topics are created and used internally by the Kafka brokers and clients.
 Kafka has several internal topics.
 These are used to store consumer offsets (`\__consumer_offsets`) or transaction state (`__transaction_state`).
 These topics can be configured using dedicated Kafka broker configuration options starting with prefix `offsets.topic.` and `transaction.state.log.`. 
--- a/books/con-message-retention.adoc
+++ b/books/con-message-retention.adoc
@@ -11,18 +11,18 @@ It can be defined based on time, partition size or both.
 
 For example, you can define that the messages should be kept:
 * For 7 days
-* Until the parition has 1GB of messages. Once the limit is reached, the oldest messages should be removed.
+* Until the parition has 1GB of messages. Once the limit is reached, the oldest messages will be removed.
 * For 7 days or until the 1GB limit has been reached.
   Whatever limit comes first will be used.
 
 WARNING: Kafka brokers store messages in log segments.
-The messages which are passed their retention policy will be deleted only when new log segment is created.
-New log segments are created when the previous log segment gets full.
+The messages which are past their retention policy will be deleted only when a new log segment is created.
+New log segments are created when the previous log segment exceeds the configured log segment size.
 Additionally, users can request new segments to be created periodically.
 
-Additionally, Kafka brokers supports compacting policy.
+Additionally, Kafka brokers supports a compacting policy.
 
-With compacting policy, the broker will always keep only the last message for each key.
+For a topic with the compacted policy, the broker will always keep only the last message for each key.
 The older messages with the same key will be removed form the partition.
 Because compacting is a periodicaly executed action, it does not happen immediately when the new message with the same key are sent to the partition. 
 Instead it might take some time until the older messages are removed.
--- a/books/con-message-retention.adoc
+++ b/books/con-message-retention.adoc
@@ -21,5 +21,8 @@ New log segments are created when the previous log segment gets full.
 Additionally, users can request new segments to be created periodically.
 
 Additionally, Kafka brokers supports compacting policy.
-With compacting policy, the broker will always try to keep the last message for given key available.
-Older messages with the same key will be deleted.
+
+With compacting policy, the broker will always keep only the last message for each key.
+The older messages with the same key will be removed form the partition.
+Because compacting is a periodicaly executed action, it doesn't happen immediately when the new message with the same key are sent to the partition. 
+Instead it might take some time until the older messages are removed.
--- a/books/con-message-retention.adoc
+++ b/books/con-message-retention.adoc
@@ -20,6 +20,6 @@ The messages which are passed their retention policy will be deleted only when n
 New log segments are created when the previous log segment gets full.
 Additionally, users can request new segments to be created periodically.
 
-Additionally, AMQ Streams supports compacting policy.
-With compacting policy, the broker will always try to keep the latest message for given key available.
+Additionally, Kafka brokers supports compacting policy.
+With compacting policy, the broker will always try to keep the last message for given key available.
 Older messages with the same key will be deleted.
--- a/books/con-message-retention.adoc
+++ b/books/con-message-retention.adoc
@@ -1,0 +1,25 @@
+// Module included in the following assemblies:
+//
+// assembly-topics.adoc
+
+[id='con-message-retention-{context}']
+
+= Message retention
+
+The message retention policy defines how long will the messages be stored on the Kafka brokers.
+It can be defined based on time, partition size or both.
+
+For example, you can define that the messages should be kept:
+* For 7 days
+* Until the parition has 1GB of messages. Once the limit is reached, the oldest messages should be removed.
+* For 7 days or until the 1GB limit has been reached.
+  Whatever limit comes first will be used.
+
+WARNING: Kafka brokers store messages in log segments.
+The messages which are passed their retention policy will be deleted only when new log segment is created.
+New log segments are created when the previous log segment gets full.
+Additionally, users can request new segments to be created periodically.
+
+Additionally, AMQ Streams supports compacting policy.
+With compacting policy, the broker will always try to keep the latest message for given key available.
+Older messages with the same key will be deleted.
--- a/books/con-message-retention.adoc
+++ b/books/con-message-retention.adoc
@@ -6,10 +6,11 @@
 
 = Message retention
 
-The message retention policy defines how long will the messages be stored on the Kafka brokers.
+The message retention policy defines how long the messages will be stored on the Kafka brokers.
 It can be defined based on time, partition size or both.
 
 For example, you can define that the messages should be kept:
+
 * For 7 days
 * Until the parition has 1GB of messages. Once the limit is reached, the oldest messages will be removed.
 * For 7 days or until the 1GB limit has been reached.
@@ -20,9 +21,11 @@ The messages which are past their retention policy will be deleted only when a n
 New log segments are created when the previous log segment exceeds the configured log segment size.
 Additionally, users can request new segments to be created periodically.
 
-Additionally, Kafka brokers supports a compacting policy.
+Additionally, Kafka brokers support a compacting policy.
 
 For a topic with the compacted policy, the broker will always keep only the last message for each key.
-The older messages with the same key will be removed form the partition.
-Because compacting is a periodicaly executed action, it does not happen immediately when the new message with the same key are sent to the partition. 
+The older messages with the same key will be removed from the partition.
+Because compacting is a periodically executed action, it does not happen immediately when the new message with the same key are sent to the partition. 
 Instead it might take some time until the older messages are removed.
+
+For more information about the message retention configuration options, see xref:ref-topic-configuration-{context}[].
--- a/books/con-message-retention.adoc
+++ b/books/con-message-retention.adoc
@@ -24,5 +24,5 @@ Additionally, Kafka brokers supports compacting policy.
 
 With compacting policy, the broker will always keep only the last message for each key.
 The older messages with the same key will be removed form the partition.
-Because compacting is a periodicaly executed action, it doesn't happen immediately when the new message with the same key are sent to the partition. 
+Because compacting is a periodicaly executed action, it does not happen immediately when the new message with the same key are sent to the partition. 
 Instead it might take some time until the older messages are removed.
--- a/books/con-partitions-and-topics.adoc
+++ b/books/con-partitions-and-topics.adoc
@@ -6,7 +6,7 @@
 
 = Partitions and replicas
 
-Messages in AMQ Streams are always sent to / recieved from a topic.
+Messages in AMQ Streams are always sent to / received from a topic.
 A topic is always split into one or more partitions.
 Partitions act as shards and every messages sent by a producer is always written only into single partition.
 Thanks to the sharding of messages into different partitions, AMQ Streams are easy to scale horizontally.
@@ -17,7 +17,7 @@ To configure the number of replicas, you have to configure the _replication fact
 _Replication factor_ defines the number of copies which will be held within the cluster.
 One of the replicas for given partition will be elected as a leader.
 The leader replica will be used by the producers to send new messages and by the consumer to consume messages.
-The other replicas will be follower replicas
+The other replicas will be follower replicas.
 The follower replicas will keep copies of the messages.
 
 NOTE: The number of replicas in the replication factor includes the leader as well as the follower replicas.
--- a/books/con-partitions-and-topics.adoc
+++ b/books/con-partitions-and-topics.adoc
@@ -6,20 +6,23 @@
 
 = Partitions and replicas
 
-Messages in Kafka are always sent to / received from a topic.
+Messages in Kafka are always sent to or received from a topic.
 A topic is always split into one or more partitions.
 Partitions act as shards.
-That means that every messages sent by a producer is always written only into a single partition.
+That means that every message sent by a producer is always written only into a single partition.
 Thanks to the sharding of messages into different partitions, topics are easy to scale horizontally.
 
 Each partition can have one or more replicas, which will be stored on different brokers in the cluster.
-The replicas will keep copies of the messages.
 When creating a topic you can configure the number of replicas using the _replication factor_.
 _Replication factor_ defines the number of copies which will be held within the cluster.
 One of the replicas for given partition will be elected as a leader.
 The leader replica will be used by the producers to send new messages and by the consumers to consume messages.
 The other replicas will be follower replicas.
-The follower replicas will keep copies of the messages.
+The followers replicate the leader.
 
-NOTE: The number of replicas in the replication factor includes the leader as well as the follower replicas.
-For example with replication factor `3`, there will be in total three replicas: one leader replica and two follower replicas.
+If the leader fails, one of the followers will automatically become the new leader.
+Each server acts as a leader for some of its partitions and a follower for others so the load is well balanced within the cluster.
+
+NOTE: 
+The replication factor determines the number of replicas including the leader and the followers.
+For example, if you set the replication factor to `3`, then there will one leader and two follower replicas.
--- a/books/con-partitions-and-topics.adoc
+++ b/books/con-partitions-and-topics.adoc
@@ -6,10 +6,10 @@
 
 = Partitions and replicas
 
-Messages in AMQ Streams are always sent to / received from a topic.
+Messages in Kafka are always sent to / received from a topic.
 A topic is always split into one or more partitions.
 Partitions act as shards and every messages sent by a producer is always written only into single partition.
-Thanks to the sharding of messages into different partitions, AMQ Streams are easy to scale horizontally.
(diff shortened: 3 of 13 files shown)


books/con-internal-topics.adoc
```diff
@@ -0,0 +1,33 @@
+// Module included in the following assemblies:
+//
+// assembly-topics.adoc
+
+[id='con-internal-topics-{context}']
+
+= Internal topics
+
+Kafka has several internal topics.
+These are used to store consumer offsets (`\__consumer_offsets`) or transaction state (`__transaction_state`).
```
ppatierno: what is this "\\" ? I guess just a typo

books/con-internal-topics.adoc
```diff
@@ -0,0 +1,33 @@
...
+The default value is `3`.
+
+`transaction.state.log.num.partitions`::
+Number of partitions for `__transaction_state` topic.
+The default value is `50`.
+
+`transaction.state.log.min.isr`::
+Minimum number of replicas that must acknowledge a write to `__transaction_state` topic to be considered successful.
+If this minimum cannot be met, then the producer will fail with an exception.
+The default value is `2`.
```
ppatierno: just asking ... if we are not listing all the possible options in our documentation, should customers go to the upstream one for the others available?

books/con-message-retention.adoc
```diff
@@ -0,0 +1,25 @@
...
+* Until the parition has 1GB of messages. Once the limit is reached, the oldest messages should be removed.
+* For 7 days or until the 1GB limit has been reached.
+  Whatever limit comes first will be used.
+
+WARNING: Kafka brokers store messages in log segments.
+The messages which are passed their retention policy will be deleted only when new log segment is created.
+New log segments are created when the previous log segment gets full.
+Additionally, users can request new segments to be created periodically.
+
+Additionally, AMQ Streams supports compacting policy.
```
ppatierno: Not sure if we have to refer as "AMQ Streams" or using "Kafka".

books/con-message-retention.adoc
```diff
@@ -0,0 +1,25 @@
...
+With compacting policy, the broker will always try to keep the latest message for given key available.
```
ppatierno: why are you using "try"? are there scenarios where it doesn't happen?

books/con-partitions-and-topics.adoc
```diff
@@ -0,0 +1,24 @@
+// Module included in the following assemblies:
+//
+// assembly-topics.adoc
+
+[id='con-partitions-and-topics-{context}']
+
+= Partitions and replicas
+
+Messages in AMQ Streams are always sent to / recieved from a topic.
```
ppatierno: typo "received"

books/con-partitions-and-topics.adoc
```diff
@@ -0,0 +1,24 @@
...
+Partitions act as shards and every messages sent by a producer is always written only into single partition.
+Thanks to the sharding of messages into different partitions, AMQ Streams are easy to scale horizontally.
+
+Each partition can exist in one or more replicas.
+The replicas will keep copies of the messages.
+To configure the number of replicas, you have to configure the _replication factor_.
+_Replication factor_ defines the number of copies which will be held within the cluster.
+One of the replicas for given partition will be elected as a leader.
+The leader replica will be used by the producers to send new messages and by the consumer to consume messages.
+The other replicas will be follower replicas
```
ppatierno: miss the full stop.

books/con-topic-autocreation.adoc
```diff
@@ -0,0 +1,17 @@
+// Module included in the following assemblies:
+//
+// assembly-topics.adoc
+
+[id='con-topic-autocreation-{context}']
+
+= Topic auto-creation
+
+When a producer or consumer tries to send or receive messages to / from a topic which does not exist, Kafka will, by default, automatically create that topic.
```
ppatierno: before you used the form "send to / receive from" now "send or receive messages to / from" ... maybe it's better using always the same?

books/con-topic-configuration.adoc
```diff
@@ -0,0 +1,60 @@
...
+The default value is `604800000` (7 days).
+
+`retention.bytes`::
+Maximum number of bytes a partition will retain.
+Once the partition size grows over this limit, the oldest log segments will be deleted.
+Value of `-1` indicates no limit.
+The default value is `-1`.
+
+`segment.bytes`::
+File size of a single commit log segment file in bytes.
```
ppatierno: shoud be add that when the size is reached a new log segment is created?

books/proc-creating-a-topic.adoc
```diff
@@ -0,0 +1,45 @@
+// Module included in the following assemblies:
+//
+// assembly-topics.adoc
+
+[id='proc-creating-a-topic-{context}']
+
+= Creating a topic
+
+The `kafka-topics.sh` tool can be used to manage topics.
+`kafka-topics.sh` is part of the AMQ Streams distribution and can be found in the `bin/` directory.
```
ppatierno: do we need the "/" at the end?

books/proc-creating-a-topic.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+.Creating a topic
+
+. Create a topic using the `kafka-topics.sh` utility.
+Specify the Zookeeper URL in the `--zookeeper` options.
+Use `--create` option to specify that a new topic should be created.
+Topic name has to be specified in the `--topic` option.
+Number of partitions and replication factor have to be specified using `--partitions` and `--replication-factor` changes.
++
+You can also override some of the default topic configuration options using the option `--config`
```
ppatierno: miss full stop.

books/proc-deleting-a-topic.adoc
```diff
@@ -0,0 +1,42 @@
+// Module included in the following assemblies:
+//
+// assembly-topics.adoc
+
+[id='proc-deleting-a-topic-{context}']
+
+= Deleting a topic
+
+The `kafka-topics.sh` tool can be used to manage topics.
+`kafka-topics.sh` is part of the AMQ Streams distribution and can be found in the `bin/` directory.
```
ppatierno: same as before about ending "/"

books/proc-describing-a-topic.adoc
```diff
@@ -0,0 +1,38 @@
+// Module included in the following assemblies:
+//
+// assembly-topics.adoc
+
+[id='proc-describing-a-topic-{context}']
+
+= Describing a topic
+
+The `kafka-topics.sh` tool can be used to manage topics.
+`kafka-topics.sh` is part of the AMQ Streams distribution and can be found in the `bin/` directory.
```
ppatierno: Ditto as before

books/proc-modifying-a-topic.adoc
```diff
@@ -0,0 +1,60 @@
+// Module included in the following assemblies:
+//
+// assembly-topics.adoc
+
+[id='proc-modifying-a-topic-{context}']
+
+= Modifying a topic
+
+The `kafka-configs.sh` tool can be used to modify topic configurations.
+`kafka-configs.sh` is part of the AMQ Streams distribution and can be found in the `bin/` directory.
```
ppatierno: Last time :-)

books/con-internal-topics.adoc
```diff
@@ -0,0 +1,33 @@
...
+//
+// assembly-topics.adoc
+
+[id='con-internal-topics-{context}']
+
+= Internal topics
+
+Kafka has several internal topics.
+These are used to store consumer offsets (`\__consumer_offsets`) or transaction state (`__transaction_state`).
+These topics can be configured using dedicated options starting with prefix `offsets.topic.` and `transaction.state.log.`.
```
tombentley: Do we call these 'options' elsewhere? I.e. is it clear enough to the reader that these things are Kafka configuration parameters?

Also, before we get into what configs these topics have, isn't it worth elaborating a little more on what "consumer offsets" are actually about, so the reader can make an informed decision about how to configure them. Ditto transaction state of course.

books/con-internal-topics.adoc
scholzj: No ... for some reason, without the `\` the `_` signs are not rendered properly nor here neither in the following topic names. Adding the `\` here solves it for all of them. Seems like some Asciidoctor issue.

I wonder if @dmuntima or @ncbaratta saw this issue already somewhere.

books/con-internal-topics.adoc
scholzj: I assume that we will not be allowed to link to Apache Kafka docs. If you think I missed any important option, I can add it here. Over time we should have our own generated list of options as Kafka does. But I assume that will not be done for beta.

books/con-message-retention.adoc
```diff
@@ -0,0 +1,25 @@
...
+[id='con-message-retention-{context}']
+
+= Message retention
+
+The message retention policy defines how long will the messages be stored on the Kafka brokers.
+It can be defined based on time, partition size or both.
+
+For example, you can define that the messages should be kept:
+* For 7 days
+* Until the parition has 1GB of messages. Once the limit is reached, the oldest messages should be removed.
```
tombentley: the oldest messages will be removed

books/con-internal-topics.adoc
ppatierno: so does it mean that you have to add the "\\" even to the following `__transaction_state` ?

books/con-message-retention.adoc
```diff
@@ -0,0 +1,25 @@
...
+* For 7 days or until the 1GB limit has been reached.
+  Whatever limit comes first will be used.
+
+WARNING: Kafka brokers store messages in log segments.
+The messages which are passed their retention policy will be deleted only when new log segment is created.
```
tombentley: past their retention policy ... only when a new log segment

books/con-message-retention.adoc
scholzj: Well, the compacting is not happening constantly. So there are situations when you might have many records with the same key which will be only compacted later.

books/con-message-retention.adoc
```diff
@@ -0,0 +1,25 @@
...
+New log segments are created when the previous log segment gets full.
```
tombentley: when the previous log segment exceeds the configured log segment size.

books/con-internal-topics.adoc
ppatierno: No, I am not saying that you missed important options. I was just wondering if we are ok listing only the most used ones or we have to list them all.

books/con-message-retention.adoc
tombentley: supports a compacting policy

books/con-message-retention.adoc
ppatierno: In my mind, it's different from using the "try" verb maybe it's just a matter of translating in Italian. Using "try" I understand that it could fail instead what you are saying is that Kafka doesn't do compaction immediately when the new message with the same key comes but it will do, for sure, later (because it's a periodic action it takes).

tombentley: For a topic with the compacted policy, the broker...

books/con-partitions-and-topics.adoc
```diff
@@ -0,0 +1,24 @@
...
+//
+// assembly-topics.adoc
+
+[id='con-partitions-and-topics-{context}']
+
+= Partitions and replicas
+
+Messages in AMQ Streams are always sent to / recieved from a topic.
+A topic is always split into one or more partitions.
+Partitions act as shards and every messages sent by a producer is always written only into single partition.
```
tombentley: "Partitions act is shards" – can we assume the reader is familiar with what a shard is?

"and every message sent"

"is written to a single partition" is enough, I think.

books/con-partitions-and-topics.adoc
```diff
@@ -0,0 +1,24 @@
...
+Thanks to the sharding of messages into different partitions, AMQ Streams are easy to scale horizontally.
```
tombentley: I would make this about the scalability of topics, rather than AMQ Streams itself.

books/con-partitions-and-topics.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+Each partition can exist in one or more replicas.
```
tombentley: We need to be careful about how we're going to use terminology here: Is a replica a copy of the partition on a different broker? Or is a replica a broker which has a copy? I think the former definition is what we should use. So: 

Each partition can have one or more replicas, which will be stored on different brokers in the cluster.

books/con-partitions-and-topics.adoc
```diff
@@ -0,0 +1,24 @@
+// Module included in the following assemblies:
+//
+// assembly-topics.adoc
+
+[id='con-partitions-and-topics-{context}']
+
+= Partitions and replicas
+
+Messages in AMQ Streams are always sent to / received from a topic.
```
ppatierno: I agree on your general comment about AMQ Streams vs Kafka usage. AMQ Streams provides more components than just Kafka and it makes sense to use AMQ Streams in a lot of places. In this case, I would use Kafka instead. We are talking about send/receive messages to/from topic which is about Kafka (so no Zookeeper involved, which is an AMQ Streams component).

books/con-partitions-and-topics.adoc
```diff
@@ -0,0 +1,24 @@
...
+A topic is always split into one or more partitions.
+Partitions act as shards and every messages sent by a producer is always written only into single partition.
+Thanks to the sharding of messages into different partitions, AMQ Streams are easy to scale horizontally.
```
ppatierno: Same as above because we are really scaling Kafka (not AMQ Streams as a whole).

books/con-partitions-and-topics.adoc
```diff
@@ -0,0 +1,24 @@
...
+= Partitions and replicas
+
+Messages in AMQ Streams are always sent to / recieved from a topic.
+A topic is always split into one or more partitions.
+Partitions act as shards and every messages sent by a producer is always written only into single partition.
+Thanks to the sharding of messages into different partitions, AMQ Streams are easy to scale horizontally.
+
+Each partition can exist in one or more replicas.
+The replicas will keep copies of the messages.
+To configure the number of replicas, you have to configure the _replication factor_.
```
tombentley: "When creating a topic you can configure the number of replicas using the _replication factor_."

(This is important, because the RF is a concept that only exists when a topic is created; Kafka doesn't internally have any such concept).

books/con-message-retention.adoc
scholzj: Thanks. Ok, I updated it based on your suggestion. I think that long term we will need to write a lot more detailed explanation of how compacting works. But this should be enough for now.

books/con-partitions-and-topics.adoc
```diff
@@ -0,0 +1,24 @@
...
+_Replication factor_ defines the number of copies which will be held within the cluster.
+One of the replicas for given partition will be elected as a leader.
```
tombentley: This is where terminology is important, because here you're using replica to mean broker-with-a-copy.

books/con-partitions-and-topics.adoc
```diff
@@ -0,0 +1,24 @@
...
+The leader replica will be used by the producers to send new messages and by the consumer to consume messages.
```
tombentley: consumers (plural)

books/con-partitions-and-topics.adoc
```diff
@@ -0,0 +1,24 @@
...
+The other replicas will be follower replicas
+The follower replicas will keep copies of the messages.
```
tombentley: The follows purpose is to keep their copy of the messages as up-to-date as possible.

books/con-topic-configuration.adoc
```diff
@@ -0,0 +1,60 @@
+// Module included in the following assemblies:
+//
+// assembly-topics.adoc
+
+[id='con-topic-configuration-{context}']
+
+= Topic configuration
+
+Auto-created topics will use the default topic configuration which can be specified in the broker properties file.
+However, when creating topics manually, their configuration can be specified at creation time.
```
tombentley: It is also possible to change a topic's configuration after it has been created.

books/con-topic-configuration.adoc
```diff
@@ -0,0 +1,60 @@
...
+Specifies the compression which is used for stored messages.
+Valid values are `gzip`, `snappy`, `lz4`, `uncompressed` (no compression) and `producer` (retain the compression codec used by the producer).
+The default value is `producer`.
+
+`max.message.bytes`::
+Maximum size of a batch of messages allowed by the Kafka broker, in bytes.
+The default value is `1000012`.
+
+`min.insync.replicas`::
+The minimum number of replicas which must be in sync for a write to be considered successful.
```
tombentley: We've not defined what in sync means

books/con-topic-configuration.adoc
tombentley: "The maximum file size..."

books/con-topic-configuration.adoc
```diff
@@ -0,0 +1,60 @@
...
+Maximum number of bytes a partition will retain.
+Once the partition size grows over this limit, the oldest log segments will be deleted.
+Value of `-1` indicates no limit.
+The default value is `-1`.
+
+`segment.bytes`::
+File size of a single commit log segment file in bytes.
+The default value is `1073741824` bytes (1 gibibyte).
+
+Similar options are available to configure in the Kafka broker the default settings for auto-created topics:
```
tombentley: The defaults for auto-created topics can be specified in the Kafka broker configuration using similar options:

books/proc-creating-a-topic.adoc
```diff
@@ -0,0 +1,45 @@
...
+`kafka-topics.sh` is part of the AMQ Streams distribution and can be found in the `bin/` directory.
+
+.Prerequisites
+
+* AMQ Streams cluster is installed and running
+
+.Creating a topic
+
+. Create a topic using the `kafka-topics.sh` utility.
+Specify the Zookeeper URL in the `--zookeeper` options.
```
tombentley: "`--zookeeper` option." (singular)

books/proc-creating-a-topic.adoc
```diff
@@ -0,0 +1,45 @@
...
+Use `--create` option to specify that a new topic should be created.
```
tombentley: Use the

books/proc-creating-a-topic.adoc
```diff
@@ -0,0 +1,45 @@
...
+Topic name has to be specified in the `--topic` option.
```
tombentley: The topic name

books/proc-creating-a-topic.adoc
```diff
@@ -0,0 +1,45 @@
...
+Number of partitions and replication factor have to be specified using `--partitions` and `--replication-factor` changes.
```
tombentley: "The number of..."

"`--partitions` and `--replication-factor` options."

books/proc-creating-a-topic.adoc
```diff
@@ -0,0 +1,45 @@
...
++
+You can also override some of the default topic configuration options using the option `--config`
+This option can be used multiple times to override different options.
++
+[source,shell,subs=+quotes]
+bin/kafka-topics.sh --zookeeper _<ZookeeperAddress>_ --create --topic _<TopicName>_ --partitions _<NumberOfPartitions>_ --replication-factor _<ReplicationFactor>_ --config _<Option1>_=_<Value1>_ --config _<Option2>_=_<Value2>_
++
+.Example of command to create a topic named `myTopic`
```
tombentley: of the command

books/proc-creating-a-topic.adoc
```diff
@@ -0,0 +1,45 @@
...
+bin/kafka-topics.sh --zookeeper zoo1.my-domain.com:2181 --create --topic myTopic --partitions 50 --replication-factor 3 --config cleanup.policy=compact --config min.insync.replicas=2
+
+. Verify that the topic exists using `kafka-topics.sh`.
++
+[source,shell,subs=+quotes]
+bin/kafka-topics.sh --zookeeper  _<ZookeeperAddress>_ --describe --topic _<TopicName>_
++
+.Example of command to describe a topic named `myTopic`
+[source,shell,subs=+quotes]
+bin/kafka-topics.sh --zookeeper zoo1.my-domain.com:2181 --describe --topic myTopic
```
tombentley: Let's avoid using camel case topic names. We never know which of these assemblies might end up being reused for AMQ Streams on OpenShift.

books/proc-modifying-a-topic.adoc
tombentley: Are we really going to include this sentence in every procedure which uses a CLI script? It's incredibly repetitive.

books/con-internal-topics.adoc
scholzj: I added there that these are _Kafka configuration options_. We are currently not explaining consumer groups and I do not expect we will have that for beta.

books/con-partitions-and-topics.adoc
scholzj: TBH, I'm not sure what you mean. Through the document I always use replica as a copy of a partition. Never as a different broker.

books/con-partitions-and-topics.adoc
scholzj: So ... any ideas how to change this if you think this is not enough?

books/con-topic-configuration.adoc
scholzj: No. And I do not expect we will have time to do that for beta. If you want I can remove this option.

books/proc-creating-a-topic.adoc
scholzj: Good point.

books/proc-modifying-a-topic.adoc
scholzj: Where would you include it then to have it only once? Its kind of part of the modular docs that the structure is sometimes a bit strange. Should we simply remove it from everywhere? It doesn't carry much information anyway.

tombentley: The thing is that we're still assembling those modules into something which might be read all the way through. It feels like what we need is a glossary, and, ideally, for the meaning to be shown only on mouseover. But what's @ncbaratta's take on this?

books/con-topic-configuration.adoc
tombentley: We'll only have to add it back again when we do have that content, so leave it for now, I think.

books/con-partitions-and-topics.adoc
tombentley: It's not really a _replica_ (as in copy) that's elected, it's a broker that has a replica that's elected. Similarly the sentence "The follower replicas will keep copies of the messages." is tautologous if you're using replica to mean copy. The way I think of it (which isn't the only interpretation) is that it's the brokers which are acting as leaders/followers for replicas. Replicas are just static, dumb things. I don't care if we use that interpretation or not, what's important is that we've defined, and use, this terminology consistently. I've not yet seen a concept which explains what these things are, did I miss that?

scholzj: I don't know. Maybe we need someone who doesn't know Kafka to read it and tell us if he understood. I think that saying that one broker is leader and others are followers is misleading. because a single broker will be leader and follower for different partitions at the same time. So this would need to be carefully worded to make sure it doesn't cause more harm.

I honestly think these things are better explained by giving these features to the replicas despite them being a fairly _virtual_ term.

books/proc-modifying-a-topic.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+.Prerequisites
+
+* AMQ Streams cluster is installed and running
+* Topic `mytopic` exists
+
+.Modify topic configuration
+
+. Use the `kafka-configs.sh` tool to get the current configuration.
+Specify the Zookeeper URL in the `--zookeeper` option.
```
dmuntima: Create a bulleted list of the options for steps 1-3. Same comment as in "Creating a topic" module.

books/proc-deleting-a-topic.adoc
```diff
@@ -0,0 +1,42 @@
...
+`kafka-topics.sh` is part of the AMQ Streams distribution and can be found in the `bin` directory.
+
+.Prerequisites
+
+* AMQ Streams cluster is installed and running
+* Topic `mytopic` exists
+
+.Deleting a topic
+
+. Delete a topic using the `kafka-topics.sh` utility.
```
dmuntima: Create a bulleted list of the options.

PR #10, opened 2018-09-25
Java program for generating config docs

tombentley:
This PR adds a small Java program to generate asciidoc documentation of the various configs. I've not hooked it into the main `Makefile`, and have for now included the generated files in the PR.

To do:

* [x] There are a few uses of HTML in the documentation which we will need to convert to asciidoc. Things like `<ul>`/`<li>` mainly, I think.
* [ ] Hook into the main Makefile, assuming this would work in the environment in which it is executed.


books/master.adoc
```diff
@@ -39,3 +39,15 @@ include::installation-kafka-producer-and-consumer.adoc[]
 include::installation-kafka-streams.adoc[]
 
 include::installation-sample-configuration.adoc[]
+
+include::ref-broker-config.adoc[leveloffset=+1]
```
scholzj: Could you move all of these above the:
```
// non-modularized content
```
line? That is where the modularized content ends.

tombentley:
Now generating topic config, and fixed the HTML issue.

tombentley:
I've had to disable the `check_docs.sh` script for the generated files since they use things which would otherwise be prohibited (e.g., i.e., ...n't). I'm going to merge this without @rh-messaging/docs review on the basis that:

1) We agree this is better than maintaining separate docs for these things ourselves
2) We therefore have little scope to change this documentation here (we should instead try to get any changes we want merged in upstream Apache Kafka).

PR #35, opened 2019-02-01
[DOCS] Create storage considerations for Streams on RHEL

laidan6000:
This pull request adds a new concept module detailing some considerations for provisioning storage for AMQ Streams on Red Hat Enterprise Linux.

### DIFF
--- a/books/con-considerations-for-data-storage.adoc
+++ b/books/con-considerations-for-data-storage.adoc
@@ -11,14 +11,14 @@ Consider the following information when choosing a data storage system for {Prod
 {ProductName} and Zookeeper both require the use of block storage.
 The use of file storage is not recommended.
 
-NOTE: {ProductName} is known to work well with cloud-based block storage solutions, including Amazon Elastic Block Store (EBS).
-
 You do not need to provision replicated storage because Kafka and Zookeeper both have built-in data replication.
 
 Choose local storage for optimal performance. If local storage is not available, you can use a Storage Area Network (SAN) accessed by a protocol such as Fibre Channel or iSCSI.
+
+NOTE: {ProductName} is known to work well with cloud-based block storage solutions, including Amazon Elastic Block Store (EBS).
 
 Although the use of solid-state drives (SSDs) is not required, they can provide improved performance in large clusters where data will be sent to and received from multiple topics asynchronously. SSDs are particularly effective for use with Zookeeper, which requires fast, low latency data access.
 
 If possible, use separate disks for data storage by Kafka and Zookeeper.
 
-It is recommended that you configure your storage system to use the XFS file system. AMQ Streams is also compatible with the ext4 file system, but this might require additional configuration.
+It is recommended that you configure your storage system to use the XFS file system. {ProductName} is also compatible with the ext4 file system, but this might require additional configuration to achieve optimal performance.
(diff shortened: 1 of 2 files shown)


books/con-considerations-for-data-storage.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+You do not need to provision replicated storage because Kafka and Zookeeper both have built-in data replication.
+
+Choose local storage for optimal performance. If local storage is not available, you can use a Storage Area Network (SAN) accessed by a protocol such as Fibre Channel or iSCSI.
+
+Although the use of solid-state drives (SSDs) is not required, they can provide improved performance in large clusters where data will be sent to and received from multiple topics asynchronously. SSDs are particularly effective for use with Zookeeper, which requires fast, low latency data access.
+
+If possible, use separate disks for data storage by Kafka and Zookeeper.
+
+It is recommended that you configure your storage system to use the XFS file system. AMQ Streams is also compatible with the ext4 file system, but this might require additional configuration.
```
scholzj:
```suggestion
It is recommended that you configure your storage system to use the XFS file system. AMQ Streams is also compatible with the ext4 file system, but this might require additional configuration to achieve optimal performance.
```

books/con-considerations-for-data-storage.adoc
```diff
@@ -0,0 +1,24 @@
...
+[id='considerations-for-data-storage-{context}']
+
+= Considerations for data storage
+
+Consider the following information when choosing a data storage system for {ProductName} on Red Hat Enterprise Linux.
+
+{ProductName} and Zookeeper both require the use of block storage.
+The use of file storage is not recommended.
+
+NOTE: {ProductName} is known to work well with cloud-based block storage solutions, including Amazon Elastic Block Store (EBS).
```
scholzj: I would probably place this on different place in the text. For example on line 19 after you talk about local storage, iSCSI and Fibre Channel. It is another implementation of block storage, so I think it would fit there better.

books/con-considerations-for-data-storage.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+You do not need to provision replicated storage because Kafka and Zookeeper both have built-in data replication.
+
+Choose local storage for optimal performance. If local storage is not available, you can use a Storage Area Network (SAN) accessed by a protocol such as Fibre Channel or iSCSI.
+
+Although the use of solid-state drives (SSDs) is not required, they can provide improved performance in large clusters where data will be sent to and received from multiple topics asynchronously. SSDs are particularly effective for use with Zookeeper, which requires fast, low latency data access.
```
scholzj:
```suggestion
Although the use of solid-state drives (SSDs) is not required for Apache Kafka, they can provide improved performance in large clusters where data will be sent to and received from multiple topics asynchronously. SSDs are particularly effective for use with Zookeeper, which requires fast, low latency data access.
```

scholzj: (we should make clear that SSDs are no required only for Kafka since that does not apply for Zookeeper)

books/con-considerations-for-data-storage.adoc
scholzj: (AFAIK ext4 works, just the performance is not optimal without the tuning)

ncbaratta: Replace AMQ Streams with a product attribute.

PR #39, opened 2019-03-06
Bump the Red Hat serial number for Maven dependencies

scholzj:
Unfortunately I had to do Kafka re-build which increases the Red Hat suffix. This PR should adapt the docs accordingly.

### DIFF
--- a/books/common/attributes.adoc
+++ b/books/common/attributes.adoc
@@ -27,7 +27,7 @@
 
 // Maven dependencies
 :MavenRepo: https://maven.repository.redhat.com/ga/
-:ArtifactVersion: 2.1.1.redhat-00001
+:ArtifactVersion: 2.1.1.redhat-00002
 
 // File locations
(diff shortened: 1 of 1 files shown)


laidan6000:
Does this information also need to be updated in the Kafka Clients and Kafka Streams assemblies?

- [Kafka clients](https://doc-stage.usersys.redhat.com/documentation/en-us/red_hat_amq/7.2/html-single/using_amq_streams_on_red_hat_enterprise_linux_rhel/#assembly-kafka-clients-str)
- [Kafka streams](https://doc-stage.usersys.redhat.com/documentation/en-us/red_hat_amq/7.2/html-single/using_amq_streams_on_red_hat_enterprise_linux_rhel/#assembly-kafka-streams-str)

scholzj:
@laidan6000 The attribute is used in these sections. So this updates these sections.

PR #40, opened 2019-03-11
[DOC] Procedure for upgrading AMQ Streams on RHEL

laidan6000:
This pull request adds a new assembly for upgrading AMQ Streams to 1.1.0, followed by upgrading the protocol versions in the recommended order.

The content is available on my file share at http://file.rdu.redhat.com/~dlaing/#assembly-upgrade-1-1-0-str


books/proc-upgrading-to-amq-streams-1-1-0.adoc
```diff
@@ -0,0 +1,56 @@
...
+NOTE: If prompted, log in to your Red Hat account. Make sure that *1.1.0* is selected in the *Version* drop-down list.
++
+. On the command line, make sure that the System Security Services Daemon (SSSD) is running.
++
+[source,shell,subs=+quotes]
+----
+sudo sssd
+----
+
+. Change the owner of the archive to the `kafka` user. In this example, the archive was downloaded to the Downloads directory.
```
scholzj: I think that in reality users will not download it to the Downloads folder because this will be running on a remote server. I wonder if we should make this a bit less desktop specific so that this doesn't look strange to the users who would probably mostly download it on one machine and copy it to the servers.

books/proc-upgrading-to-amq-streams-1-1-0.adoc
```diff
@@ -0,0 +1,56 @@
...
+.Prerequisites
+* {ProductName} 1.0 is installed. For instructions, see xref:proc-installing-amq-streams-{context}[Installing {ProductName}].
+
+.Procedure
+
+. Download the *Red Hat AMQ Streams 1.1.0* archive from the {ReleaseDownload110} area of the Customer Portal.
++
+NOTE: If prompted, log in to your Red Hat account. Make sure that *1.1.0* is selected in the *Version* drop-down list.
++
+. On the command line, make sure that the System Security Services Daemon (SSSD) is running.
```
scholzj: Do we really need this? I don't think we talked about `sssd` at all in the original installation guide and I don't remember anyone complaining about it.

books/assembly-upgrade-1-1-0.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+= Upgrading an {ProductName} cluster from 1.0.0 to 1.1.0
+
+This chapter describes how to upgrade {ProductName} on Red Hat Enterprise Linux to version 1.1.0, which is based on Apache Kafka 2.1.1. No cluster downtime is required. 
+
+Although AMQ Streams 1.1.0 and Kafka 2.1.1 are both minor releases, the Kafka protocol has changed since Kafka 2.0.0 was released (in particular, the message format version and inter-broker protocol version are both now at version 2.1). Therefore, the upgrade process involves configuration changes to your existing Kafka brokers and code changes to client applications (consumers and producers).
+
+Upgrading to {ProductName} 1.1.0 is a two-stage process:
+
+. First, you upgrade your {ProductName} installation to version 1.1.0, and then conduct performance testing.
```
scholzj: I think we need to clarify which parts need to be done on all machines and which parts should be done machine by machine. 

For example here if the user goes and upgrades the installation on all servers and only then goes and changes the config for the protocol versions, every restart will automatically load the new version.

So afaik the best approach is either to join some of these steps or change the order (i.e. first set the message version and interbroker protocol to 2.0 on all brokers and afterwards install the software on all brokers etc.)

scholzj: I'm not sure I understand what do they need to _performance test_. I think we should leave it out or if it is something we really need, we should explain how to do it.

books/assembly-upgrade-1-1-0.adoc
```diff
@@ -0,0 +1,40 @@
...
+//
+// master.adoc
+
+[id='assembly-upgrade-1-1-0-{context}']
+
+= Upgrading an {ProductName} cluster from 1.0.0 to 1.1.0
+
+This chapter describes how to upgrade {ProductName} on Red Hat Enterprise Linux to version 1.1.0, which is based on Apache Kafka 2.1.1. No cluster downtime is required. 
+
+Although AMQ Streams 1.1.0 and Kafka 2.1.1 are both minor releases, the Kafka protocol has changed since Kafka 2.0.0 was released (in particular, the message format version and inter-broker protocol version are both now at version 2.1). Therefore, the upgrade process involves configuration changes to your existing Kafka brokers and code changes to client applications (consumers and producers).
```
scholzj: Should we mention Zookeeper? Zookeeper version haven't changed, but they should probably still get the new binaries.

books/con-kafka-versions-overview.adoc
```diff
@@ -0,0 +1,16 @@
...
+[id='con-kafka-versions-overview-{context}']
+
+= Kafka versions overview
+
+After you have upgraded to {ProductName} 1.1.0, you can upgrade all Kafka brokers and client applications (consumers and producers) to Kafka 2.1.1. As shown in the following table, Kafka 2.1.1 uses a higher inter-broker protocol version and log message format version than Kafka 2.0.0.
+
+[options="header"]
+|=======================
+|{ProductName} version |Kafka version |Inter-broker protocol version  |Log message format version | Zookeeper version
+|1.0                   |2.0.0         |2.0                           |2.0                        | 3.4.13
```
scholzj: Would it be better to list both as `x.y.z` or both as `x.y`? I found it a bit strange that we have `1.0` but `1.1.0`.

books/proc-downgrading-kafka-brokers-to-previous-protocol-versions.adoc
```diff
@@ -0,0 +1,42 @@
+// Module included in the following assemblies:
+//
+// assembly-upgrade-1-1-0.adoc
+
+[id='proc-downgrading-kafka-brokers-to-previous-protocol-versions-{context}']
+
+= Downgrading Kafka brokers to the previous protocol versions
```
scholzj: I do not think we are really downgrading anything here. We are just fixing the protocol version to the one of Kafka 2.0 / AMQ Streams 1.0. _Downgrading_ is imho hot the right word.

books/proc-downgrading-kafka-brokers-to-previous-protocol-versions.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+Manually configure and restart all the Kafka brokers in your cluster to use the previous inter-broker protocol and message format versions (2.0), and then test performance.
```
scholzj: Again, _test performance_ sounds strange here ... why and if we keep it how? IMHO I would just remove it.

books/proc-downgrading-kafka-brokers-to-previous-protocol-versions.adoc
```diff
@@ -0,0 +1,42 @@
...
+//
+// assembly-upgrade-1-1-0.adoc
+
+[id='proc-downgrading-kafka-brokers-to-previous-protocol-versions-{context}']
+
+= Downgrading Kafka brokers to the previous protocol versions
+
+Manually configure and restart all the Kafka brokers in your cluster to use the previous inter-broker protocol and message format versions (2.0), and then test performance. 
+
+NOTE: To avoid cluster downtime, update and restart the Kafka brokers one-by-one.
```
scholzj: Should we add at least some pointers to how to find out that the restarted broker is _up to date_ again after the restart?

books/proc-downgrading-kafka-brokers-to-previous-protocol-versions.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+= Downgrading Kafka brokers to the previous protocol versions
+
+Manually configure and restart all the Kafka brokers in your cluster to use the previous inter-broker protocol and message format versions (2.0), and then test performance. 
+
+NOTE: To avoid cluster downtime, update and restart the Kafka brokers one-by-one. 
+
+.Prerequisites
+
+* You have upgraded to {ProductName} 1.1.0.
```
scholzj: As suggested before - I think this step should be done before the 1.1 installation. Otherwise random restart will trigger 2.1.1 with the default protocol.

books/proc-downgrading-kafka-brokers-to-previous-protocol-versions.adoc
```diff
@@ -0,0 +1,42 @@
...
++
+[source,shell,subs=+quotes]
+----
+inter.broker.protocol.version=2.0
+log.message.format.version=2.0
+----
++
+This configures the Kafka broker to process data using the previous inter-broker protocol and message format versions.
+
+. On the command line, restart the Kafka broker:
```
scholzj: We should probably add a shutdown step before this. You have to stop the old broker before starting the new one.

books/proc-downgrading-kafka-brokers-to-previous-protocol-versions.adoc
```diff
@@ -0,0 +1,42 @@
...
++
+[source,shell,subs=+quotes]
+----
+/opt/kafka/bin/kafka-server-start.sh -daemon /opt/kafka/config/server.properties
+----
++
+NOTE: The broker will start using the version 1.1.0 JAR files.
```
scholzj:
```suggestion
NOTE: The broker will start using the version 1.1.0 binaries.
```

books/proc-downgrading-kafka-brokers-to-previous-protocol-versions.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+. Repeat steps 1 through 3 for all Kafka brokers in your cluster.
+
+. When all Kafka brokers are updated, test the performance of the cluster.
```
scholzj: As always ... this should probably not be here.

books/proc-updating-kafka-brokers-to-new-inter-broker-protocol-version.adoc
```diff
@@ -0,0 +1,32 @@
+// Module included in the following assemblies:
+//
+// assembly-upgrade-1-1-0.adoc
+
+[id='proc-updating-kafka-brokers-to-new-inter-broker-protocol-version-{context}']
+
+= Updating Kafka brokers to the new inter-broker protocol version
+
+If cluster performance testing was successful, manually configure and restart all the Kafka brokers to use the new inter-broker protocol version (2.1). After performing these steps, data will be transmitted between the Kafka brokers using inter-broker protocol version 2.1. Messages received are still appended to the message logs in message format version 2.0.
```
scholzj: Performance testing again.

books/proc-updating-kafka-brokers-to-new-inter-broker-protocol-version.adoc
```diff
@@ -0,0 +1,32 @@
...
+//
+// assembly-upgrade-1-1-0.adoc
+
+[id='proc-updating-kafka-brokers-to-new-inter-broker-protocol-version-{context}']
+
+= Updating Kafka brokers to the new inter-broker protocol version
+
+If cluster performance testing was successful, manually configure and restart all the Kafka brokers to use the new inter-broker protocol version (2.1). After performing these steps, data will be transmitted between the Kafka brokers using inter-broker protocol version 2.1. Messages received are still appended to the message logs in message format version 2.0. 
+
+WARNING: Downgrading to {ProductName} 1.0.0 is not possible after completing this procedure.
```
scholzj: @tombentley is this true? I would expect that the point of not going back would not be the inter-broker message protocol but the message format.

books/proc-updating-kafka-brokers-to-new-inter-broker-protocol-version.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+NOTE: To avoid downtime, update and restart the Kafka brokers one-by-one.
+
+.Prerequisites
+* You have upgraded to {ProductName} 1.1.0.
+* You have downgraded Kafka brokers to use the previous inter-broker protocol and message format versions (2.0) and tested performance.
```
scholzj: _downgreaded_ => _fixed_ or something like that.

books/proc-updating-kafka-brokers-to-new-inter-broker-protocol-version.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+.Procedure
+
+. In a text editor, open the broker properties file for the Kafka broker you want to update first. These files are commonly stored in the `/opt/kafka/config/` directory.
+
+. Set the `inter.broker.protocol.version` to `2.1`.
+
+. On the command line, restart the Kafka broker:
```
scholzj: Thsi is not a restart command. We need to add a shutdown command before this.

books/proc-upgrading-clients-to-new-kafka-version.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+.Prerequisites
+
+* You have upgraded to {ProductName} 1.1.0.
+* You have downgraded Kafka brokers to use the previous inter-broker protocol and message format versions (2.0) and tested performance.
+* You have updated Kafka brokers to use the new inter-broker protocol version (2.1).
+
+.Procedure
+
+. Upgrade all consumers to use version 1.1.0 of the {ProductName} client libraries.
```
scholzj: Since they might be using different clients ... should we rather talk about Kafka 2.1.1 clients?

books/proc-upgrading-to-amq-streams-1-1-0.adoc
```diff
@@ -0,0 +1,56 @@
...
+// assembly-upgrade-1-1-0.adoc
+
+[id='proc-upgrading-to-amq-streams-1-1-0-{context}']
+
+= Upgrading to {ProductName} 1.1.0
+
+To upgrade your cluster to {ProductName} 1.1.0, download and extract the *Red Hat AMQ Streams 1.1.0* archive from the Red Hat Customer Portal. 
+
+.Prerequisites
+* {ProductName} 1.0 is installed. For instructions, see xref:proc-installing-amq-streams-{context}[Installing {ProductName}].
```
scholzj: Should we explain that users who did not install 1.0 should just follow the guide and install 1.1 directly? Or do you think that will be clear?

books/proc-upgrading-to-amq-streams-1-1-0.adoc
```diff
@@ -0,0 +1,56 @@
...
+----
+
+. Move the archive to the `/opt/kafka` directory that was created during the installation of {ProductName} 1.0.
++
+[source,shell,subs=+quotes]
+----
+sudo mv amq-streams-1.1.0-bin.zip /opt/kafka
+----
+
+. Change the current user ID to the `kafka` user and enter the password.
```
scholzj: The prerequisite of the assembly is that user is logged as `kafka` user. Here we ask the user to sudo to `kafka` user again ... Isn't that confusing?

books/proc-updating-kafka-brokers-to-new-inter-broker-protocol-version.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+[id='proc-updating-kafka-brokers-to-new-inter-broker-protocol-version-{context}']
+
+= Updating Kafka brokers to the new inter-broker protocol version
+
+If cluster performance testing was successful, manually configure and restart all the Kafka brokers to use the new inter-broker protocol version (2.1). After performing these steps, data will be transmitted between the Kafka brokers using inter-broker protocol version 2.1. Messages received are still appended to the message logs in message format version 2.0. 
+
+WARNING: Downgrading to {ProductName} 1.0.0 is not possible after completing this procedure.
+
+NOTE: To avoid downtime, update and restart the Kafka brokers one-by-one.
```
tombentley: It's not really as simple as this. It is possible to roll the brokers too quickly which can cause problems for the clients. For example if a topic has `min.insync.replicas` and producers have `acks=all`/`acks=-1` then producers could observe a loss of availability if you restarted broker X and before in was in sync you restarted broker Y. Similarly the user can think they've restarted a broker but maybe they hard killed it and so it has to do log recovery (which can take a while) meanwhile the user is busy rolling other brokers.

The long term fix for this is to write a procedure for how to do a graceful rolling restart (i.e. describing what to look for in the logs to know that the broker is _really_ back up and functioning correctly within the cluster), and to reference that procedure when specific procedures require a rolling restart. Otherwise you will end up writing lots of procedures which each describe how to do a graceful rolling restart and it will be a maintenance nightmare. Or maybe use a snippet.

books/proc-upgrading-clients-to-new-kafka-version.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+. Optional: To avoid a temporary reduction in the performance of the cluster, perform the following steps. Otherwise, go to step three.
+
+.. Set the message format version on a topic-by-topic basis for each producer. This avoids messages received by the Kafka brokers from being converted down to message format version 2.0, which is likely to increase CPU usage and reduce cluster performance.
```
tombentley: `message.format.version` is a topic config, not a producer config, so what @laidan6000 has described does not work. The only way to avoid downconversion _completely_ is to:

* set the topic's `message.format.version` to the new version
* upgrade all the consumers for the topic before any of the producers
* upgrade the producers.

But this can be a bit of a big bang approach in that if upgrading the producer causes problems somehow then by the time you know it you're already committed to the new version.

The reality is that downconversion isn't entirely bad except when the perf impact is a problem. The best way to mitigate that is by upgrading clients on a topic-by-topic basis, so that downconversion is only happening on a few topics at a time. Obviously in real life the user would typically pick on the least important, lower throughput topics first and upgrade those and progress to the more valuable and higher throughput topics once they've proved that their own procedures for upgrading their client applications to 1.1 are working OK.

To be honest it might be worth having a short concept to explain this, because really it's up to the user to decide the best strategy which will work for them in rolling out the new client applications.

books/proc-upgrading-clients-to-new-kafka-version.adoc
```diff
@@ -0,0 +1,38 @@
...
+bin/kafka-configs.sh --zookeeper _<ZookeeperAddress>_ --entity-type topics --entity-name <TopicName> --alter --add-config message.format.version=2.1
+----
+
+.. Identify the producer or producers that write data to the topic you modified in step 2a, and then upgrade them to use message format version 2.1.
+
+.. Repeat steps a and b for all topics in your cluster, and for all producers.
+
+. Optional: If you did not complete step two, upgrade all producers to use message format version 2.1.
++
+NOTE: Unless you performed step two, messages sent by producers to topics are now converted down to message format version 2.0 before being appended to the message logs. This is likely to cause a reduction in the performance of the cluster. To restore normal performance, update all Kafka brokers to use message format version 2.0 as soon as possible. For details, see xref:proc-updating-kafka-brokers-to-new-message-format-version-{context}[Updating Kafka brokers to the new message format version].
```
tombentley: Be explicit about what is performing this conversion (the brokers).

See what I wrote above about the `message.format.version` being a topic level config. To restore normal perf you need to ensure the broker doesn't have to downconvert. Until the producer is upgraded that's not a problem. Once it is upgraded then downconversion will happen while `message.format.version=<old version>`, and you can only change it once all the consumers are using the new version of the client library. That's why upgrading the consumers first is helpful, as it means you're in the position to change `message.format.version` almost straight away (or even before you upgrade the producer if you're feeling brave).

books/proc-updating-kafka-brokers-to-new-message-format-version.adoc
```diff
@@ -0,0 +1,32 @@
+// Module included in the following assemblies:
+//
+// assembly-upgrade-1-1-0.adoc
+
+[id='proc-updating-kafka-brokers-to-new-message-format-version-{context}']
+
+= Updating Kafka brokers to the new message format version
+
+When client applications have been upgraded, you can update the Kafka brokers to use the new message format version (2.1).
```
tombentley: I think the point that users need to understand here is that the brokers `log.message.format.version` is the default value for the topic-level `message.format.version`. So if they have followed the previous procedure for upgrading their clients on a topic-by-topic basis then all their existing topics will have `message.format.version` set to the new version. In that case this proc doesn't really do anything (though it would affect new topics created without `message.format.version` set).

On the other hand, you cannot assume that all users will upgrade their clients following that advice. It might be easier for them to just upgrade all the consumers and all the producers, without bothering to change topic configs. At that point their brokers are having to do a lot of downconversion and this procedure is really important for having the brokers stop doing that.

books/proc-updating-kafka-brokers-to-new-message-format-version.adoc
```diff
@@ -0,0 +1,32 @@
...
+//
+// assembly-upgrade-1-1-0.adoc
+
+[id='proc-updating-kafka-brokers-to-new-message-format-version-{context}']
+
+= Updating Kafka brokers to the new message format version
+
+When client applications have been upgraded, you can update the Kafka brokers to use the new message format version (2.1). 
+
+NOTE: To avoid downtime, update and restart the Kafka brokers one-by-one.
```
tombentley: Same comment here about restarting brokers too quickly.

books/proc-updating-kafka-brokers-to-new-message-format-version.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+.Prerequisites
+* You have upgraded to {ProductName} 1.1.0
+* You have downgraded Kafka brokers to use the previous inter-broker protocol and message format versions and tested performance.
```
tombentley: I agree with Jakub that using "downgraded" to describe _configuring_ these things is a little misleading.

books/proc-updating-kafka-brokers-to-new-message-format-version.adoc
```diff
@@ -0,0 +1,32 @@
...
+* You have updated Kafka brokers to use the new inter-broker protocol version.
+* You have upgraded client applications to the new Kafka version.
```
tombentley: I think they key thing here is that they've upgraded all consuming applications which are consuming from topics which don't have `message.format.version` configured. (If they've not upgraded all the producers that shouldn't matter).

books/proc-downgrading-kafka-brokers-to-previous-protocol-versions.adoc
laidan6000: OK. I've reordered those two procedures. Pull request to follow.

books/proc-upgrading-to-amq-streams-1-1-0.adoc
laidan6000: I needed to enable SSSD on my machine to get it to work, but I suppose that's not representative. I'll remove the step.

laidan6000:
I've addressed the review comments. Please check this over again, thank you.

Summary of changes:

- Added a new procedure _Updating the Zookeeper binaries_.

- Reorganised content in the main assembly - brought Kafka versions table up and divided it into two

- Combined the "3 stages" list and the outline TOC.

- State that AMQ Streams 1.0.0 was based on Kafka 2.0.0

- Changed the order of the procedures. Setting protocol versions to 2.0 on all clusters is now done before installing Streams 1.1.0 on all brokers.

- Removed all references to performance testing.

- Moved the Kafka user prereq from assembly to topic level. Moved 1.0 installation to the assembly prereqs.

- Changed "downgrading" to "updating"

- Added optional step to stop the most recent Kafka broker to pdating Kafka brokers to the previous protocol versions

- Various other review comments.

- Added Tom's new concept on Upgrade strategies for client applications.

- Modified the client upgrade procedure to document the topic-by-topic strategy.

books/assembly-overview.adoc
```diff
@@ -15,10 +15,10 @@ It consists of the following main components:
 Zookeeper:: Service for highly reliable distributed coordination.
 Kafka Broker:: Messaging broker responsible for delivering records from producing clients to consuming clients.
 Kafka Connect:: A toolkit for streaming data between Kafka brokers and other systems using _Connector_ plugins.
-Kafka Consumer and Producer APIs:: Java based APIs for producing and consuming messages to / from Kafka brokers.
+Kafka Consumer and Producer APIs:: Java based APIs for producing and consuming messages to and from Kafka brokers.
 Kafka Streams API:: API for writing _stream processor_ applications.
 
-Kafka Broker is the central point connecting all these components. 
+The Kafka broker is the central point connecting all these components.
```
tombentley: I know it's talking abstractly about "the Kafka broker", but I nonetheless think this is a little misleading because it's talking about a singular broker, but in reality it's a cluster. 

```suggestion
A cluster of Kafka brokers is the hub connecting all these components. 
```

books/assembly-upgrade-1-1-0.adoc
```diff
@@ -0,0 +1,87 @@
...
+|Kafka version |Inter-broker protocol version |Log message format version |Zookeeper version
+|2.0.0         |2.0                           |2.0                        |3.4.13
+|2.1.1         |2.1                           |2.1                        |3.4.13
+|=======================
+
+Although Kafka 2.0.0 and 2.1.1 use the same version of Zookeeper, Red Hat recommends that you update your Zookeeper cluster to use the newest Zookeeper binaries before proceeding with the main {ProductName} 1.1.0 upgrade.
+
+Message format version::
+
+In Kafka, there are two different methods for setting the message format version.
```
tombentley: I think you need a sentence just to explain what this is.

"""
When a Kafka producer sends a message to a broker that message is encoded using a specific format. The broker can be configured to convert messages from newer format verions to a given older format version before the message is appended to the log.
"""

books/assembly-upgrade-1-1-0.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+Procedure outline::
+
+In summary, upgrading to {ProductName} 1.1.0 is a four-stage process:
+
+. First, you update your Zookeeper cluster to use the newest Zookeeper binaries.
+
+.. xref:proc-updating-zookeeper-binaries-{context}[Updating the Zookeeper binaries]
+
+. Second, you update all Kafka brokers to use the previous protocol versions.
```
tombentley:
```suggestion
. Second, you configure all Kafka brokers to use the previous protocol versions.
```

books/assembly-upgrade-1-1-0.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+.. xref:proc-updating-kafka-brokers-to-new-inter-broker-protocol-version-{context}[Updating Kafka brokers to the new inter-broker protocol version] 
+
+.. Decide on an appropriate strategy for upgrading your client applications based on xref:con-strategies-for-upgrading-clients-{context}[Strategies for upgrading clients].
+
+.. xref:proc-upgrading-clients-to-new-kafka-version-{context}[Upgrading client applications to the new Kafka version] (describes one upgrade strategy in more detail).
+
+.. xref:proc-updating-kafka-brokers-to-new-message-format-version-{context}[Updating Kafka brokers to the new message format version]
+
+Red Hat recommends that you complete the upgrade procedures in the order shown.
```
tombentley: Really this is more than a recommendation. If they were to do things in a different order things would likely break. So we should put it more forcefully.

```suggestion
In order to upgrade brokers and clients without downtime you need to complete the upgrade procedures in the order given above.
```

books/proc-updating-zookeeper-binaries.adoc
```diff
@@ -0,0 +1,25 @@
...
+//
+// assembly-upgrade-1-1-0.adoc
+
+[id='proc-updating-zookeeper-binaries-{context}']
+
+= Updating the Zookeeper binaries
+
+Before upgrading to {ProductName} 1.1.0, update your Zookeeper cluster with the latest binaries.
+
+. Download the latest stable release of Zookeeper from the link:http://zookeeper.apache.org/releases.html[Apache Zookeeper Releases] page.
```
tombentley: This is wrong. We do not expect people to go downloading Zookeeper from Apache, it is packaged in the AMQ Streams download.

books/proc-updating-kafka-brokers-to-previous-protocol-versions.adoc
```diff
@@ -0,0 +1,51 @@
...
+//
+// assembly-upgrade-1-1-0.adoc
+
+[id='proc-updating-kafka-brokers-to-previous-protocol-versions-{context}']
+
+= Updating Kafka brokers to the previous protocol versions
+
+Manually configure and restart all the Kafka brokers in your cluster to use the previous inter-broker protocol and message format versions (2.0). These are the versions that were used in {ProductName} 1.0 and Kafka 2.0.
+
+NOTE: Update and restart the Kafka brokers one-by-one. Before you restart a modified broker, stop the broker you configured and restarted previously.
```
tombentley: I don't understand this:

> Before you restart a modified broker, stop the broker you configured and restarted previously.

books/proc-updating-kafka-brokers-to-previous-protocol-versions.adoc
```diff
@@ -0,0 +1,51 @@
...
+.Prerequisites
+
+* You have updated the Zookeeper binaries.
+* You are logged in to Red Hat Enterprise Linux as the `kafka` user.
+
+.Procedure
+
+. In a text editor, open the broker properties file for the Kafka broker you want to update. Broker properties files are commonly stored in the `/opt/kafka/config/` directory.
+
+. Temporarily override the default inter-broker protocol and message format versions for Kafka 2.1.1 by adding the following new properties to the file:
```
tombentley: Strictly speaking these might already exist in the file for some reason.

```suggestion
. Temporarily override the default inter-broker protocol and message format versions for Kafka 2.1.1 by adding or updating the following properties to the file:
```

books/proc-updating-kafka-brokers-to-previous-protocol-versions.adoc
```diff
@@ -0,0 +1,51 @@
...
++
+[source,shell,subs=+quotes]
+----
+inter.broker.protocol.version=2.0
+log.message.format.version=2.0
+----
++
+This configures the Kafka broker to process data using the previous inter-broker protocol and message format versions.
+
+. On the command line, stop the Kafka broker that you most recently modified and restarted as part of this procedure. If you are modifying the first Kafka broker of this procedure, go to step four.
```
tombentley: I don't understand this. The user would edit the Kafka config of a particular broker and then stop and restart that same broker. Typically both these operations would be done while logged into whatever server has the broker running on it. In other words they:

1. Log in over `ssh`
2. Edit the file.
3. Stop the broker.
4. Restart the broker.
5. Exit that `ssh` session.

And repeat on each server.

laidan6000:
My file share is now up-to-date with the latest version - http://file.rdu.redhat.com/~dlaing/#assembly-upgrade-1-1-0-str

To check that you're seeing the latest version, look for the new "Strategies for upgrading clients" concept.

books/proc-upgrading-kafka-brokers-to-amq-streams-1-1-0.adoc
```diff
@@ -0,0 +1,43 @@
...
+----
+sudo mv /tmp/kafka-1-1-0/kafka_y.y-x.x.x/* /opt/kafka/
+----
+
+. Delete the temporary directory.
++
+[source,shell,subs=+quotes]
+----
+rm -r /tmp/kafka-1-1-0
+----
```
tombentley: It looks like you got confused between this procedure and proc-updating-kafka-brokers-to-previous-protocol-versions.adoc. I think they can be merged together. The algorithm is:

```
for each broker in the cluster:
    1. install the new software (as described here)
    2. update its server.properties to add log.message.format.version and inter.broker.protocol.version (as described in proc-updating-kafka-brokers-to-previous-protocol-versions.adoc)
    3. Stop the broker
    4. Start the broker
    5. Check the broker has caught up.
```

By doing both the server.properties update and the installation of the new software you avoid needing two rolling updates of the cluster.

books/proc-upgrading-clients-to-new-kafka-version.adoc
```diff
@@ -0,0 +1,38 @@
...
+[id='proc-upgrading-clients-to-new-kafka-version-{context}']
+
+= Upgrading client applications to the new Kafka version
+
+Upgrade consumers and producers to use {ProductName} 1.1.0. Red Hat recommends that you upgrade consumers first, as described in the following procedure.
+
+.Prerequisites
+
+* You have upgraded to {ProductName} 1.1.0.
+* You have downgraded Kafka brokers to use the previous inter-broker protocol and message format versions (2.0) and tested performance.
```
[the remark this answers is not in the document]
laidan6000: This isn't the latest version of the file. All references to performance testing have been removed.

books/con-strategies-for-upgrading-clients.adoc
```diff
@@ -0,0 +1,63 @@
...
+This is because brokers perform down-conversion when they receive messages from producers, not when they are sent to consumers.
+
+There are a number of strategies you can use to upgrade your clients:
+
+Consumers first::
+. Upgrade all the consuming applications.
+. Change the broker-level `log.message.format.version` to the new version.
+. Upgrade all the producing applications.
++
+This strategy is straightforward, and avoids any broker down-conversion. However, it assumes that all consumers in your organization can be upgraded in a coordinated way, and it doesn't work for applications that are both consumers and producers. There is also a risk that, if there is a problem with the upgraded clients, new-format messages might get added to the message log.
```
tombentley:
```suggestion
This strategy is straightforward, and avoids any broker down-conversion.
However, it assumes that all consumers in your organization can be upgraded in a coordinated way, and it doesn't work for applications that are both consumers and producers.
There is also a risk that, if there is a problem with the upgraded clients, new-format messages might get added to the message log so that you cannot revert to the previous consumer version.
```

books/proc-upgrading-clients-to-new-kafka-version.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+. Verify that the upgraded applications function correctly.
+
+. Change the topic's `message.format.version` configuration option to `2.1`.
++
+[source,shell,subs=+quotes]
+----
+bin/kafka-configs.sh --zookeeper _<ZookeeperAddress>_ --entity-type topics --entity-name <TopicName> --alter --add-config message.format.version=2.1
+----
+
```
tombentley: Do we need a final step to say continue upgrading remaining clients and topics?

books/proc-updating-kafka-brokers-to-new-message-format-version.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+NOTE: Update and restart the Kafka brokers one-by-one. Before you restart a modified broker, stop the broker you configured and restarted previously.
+
+.Prerequisites
+
+* You have updated the Zookeeper binaries.
+* You have updated all Kafka brokers to use the previous inter-broker protocol and message format versions (2.0).
+* You have upgraded all Kafka brokers to {ProductName} 1.1.0
+* You have updated Kafka brokers to use the new inter-broker protocol version (2.1).
+* You have upgraded supported client applications that consume messages from topics for which the `message.format.version` property is _not_ manually configured.
```
tombentley:
```suggestion
* You have upgraded supported client applications that consume messages from topics for which the `message.format.version` property is _not_ explicitly configured at the topic level
```

books/assembly-upgrade-1-1-0.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+The remainder of this chapter is divided into six separate procedures; Red Hat recommends that you complete them in the order shown.
+
+. xref:proc-upgrading-to-amq-streams-1-1-0-{context}[Upgrading to {ProductName} 1.1.0]
+. xref:proc-downgrading-kafka-brokers-to-previous-protocol-versions-{context}[Downgrading Kafka brokers to the previous protocol versions]
+. xref:proc-updating-kafka-brokers-to-new-inter-broker-protocol-version-{context}[Updating Kafka brokers to the new inter-broker protocol version]
+. xref:proc-upgrading-clients-to-new-kafka-version-{context}[Upgrading client applications to the new Kafka version]
+. xref:proc-updating-kafka-brokers-to-new-message-format-version-{context}[Updating Kafka brokers to the new message format version]
+
+.Prerequisites
```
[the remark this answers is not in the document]
seanhuck: Is "you" necessary in these statements? Perhaps simplify to " First, update your Zookeeper cluster..." for example.

books/con-strategies-for-upgrading-clients.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+Consuming applications need to receive messages in a message format that they understand. You can ensure that this is the case in one of two ways:
+
+* By upgrading all the consumers for a topic _before_ upgrading any of the producers.
+* By having the brokers down-convert messages to an older format.
+
+Using broker down-conversion puts extra load on the brokers, so it is not ideal to rely on down-conversion for all topics for a prolonged period of time. 
+For brokers to perform optimally they should not be down converting messages at all.
+
+Broker down-conversion is configured in two ways.
```
seanhuck: Needs a colon at the end.

books/con-strategies-for-upgrading-clients.adoc
```diff
@@ -0,0 +1,65 @@
...
+Per-topic consumers first, with down conversion::
+For each topic:
++
+. Change the topic-level `message.format.version` to the old version
+(or rely on the topic defaulting to the broker-level `log.message.format.version`).
+. Upgrade all the consuming and producing applications.
+. Verify that the upgraded applications function correctly.
+. Change the topic-level `message.format.version` to the new version.
++
+This strategy requires broker down-conversion, but the load on the brokers is minimized because it is only required for a single topic (or small group of topics) at a time. It also works for applications that are both consumers and producers of the same topic. This approach gives you certainty that the upgraded producers and consumers are working correctly before you commit to using the new message format version.
```
seanhuck: consider replacing "gives you certainty" with "ensures".

books/proc-updating-kafka-brokers-to-previous-protocol-versions.adoc
seanhuck: +1. Please reword to be more specific.

books/assembly-upgrade-1-1-0.adoc
laidan6000: Resolved

books/proc-updating-zookeeper-binaries.adoc
laidan6000: @PaulRMellor

books/proc-updating-kafka-brokers-to-previous-protocol-versions.adoc
laidan6000: @PaulRMellor - I added the "for all brokers" statement to the start of the procedure.

books/proc-updating-kafka-brokers-to-previous-protocol-versions.adoc
laidan6000: Combined this topic with proc-upgrading-kafka-brokers-to-amq-streams-1-1-0

books/con-strategies-for-upgrading-clients.adoc
laidan6000: @PaulRMellor

books/con-strategies-for-upgrading-clients.adoc
laidan6000: @PaulRMellor

laidan6000:
@PaulRMellor to pick up the remaining comments.

books/assembly-upgrade-1-1-0.adoc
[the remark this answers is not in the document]
PaulRMellor: Simplified and removed the 'you'

books/con-strategies-for-upgrading-clients.adoc
PaulRMellor: Updated

books/con-strategies-for-upgrading-clients.adoc
PaulRMellor: Replaced

books/proc-updating-zookeeper-binaries.adoc
PaulRMellor: Removed the statement and reworked the steps

PR #48, opened 2019-04-16
[DOCS] Kafka Connect and Mirror Maker upgrades, plus minor style edits

laidan6000:
This pull request clarifies that Kafka Connect and Mirror Maker are just client applications and should be upgraded to the new Kafka version along with other clients. It also makes a few style and wording changes. 

You can view the staged documentation on my file share: http://file.rdu.redhat.com/~dlaing/#assembly-upgrade-1-1-0-str

### DIFF
--- a/books/assembly-upgrade-1-1-0.adoc
+++ b/books/assembly-upgrade-1-1-0.adoc
@@ -26,7 +26,7 @@ Although {ProductName} 1.1.0 and Kafka 2.1.1 are both minor releases, the Kafka 
 |2.1.1         |2.1                           |2.1                        |3.4.13
 |=======================
 
-Although Kafka 2.0.0 and 2.1.1 use the same version of Zookeeper, Red Hat recommends that you update your Zookeeper cluster to use the newest Zookeeper binaries before proceeding with the main {ProductName} 1.1.0 upgrade.
+Although Kafka 2.0.0 and 2.1.1 use the same version of Zookeeper, it is recommended that you update your Zookeeper cluster to use the newest Zookeeper binaries before proceeding with the main {ProductName} 1.1.0 upgrade.
 
 == Message format version
 
--- a/books/assembly-upgrade-1-1-0.adoc
+++ b/books/assembly-upgrade-1-1-0.adoc
@@ -28,7 +28,7 @@ Although {ProductName} 1.1.0 and Kafka 2.1.1 are both minor releases, the Kafka 
 
 Although Kafka 2.0.0 and 2.1.1 use the same version of Zookeeper, Red Hat recommends that you update your Zookeeper cluster to use the newest Zookeeper binaries before proceeding with the main {ProductName} 1.1.0 upgrade.
 
-Message format version::
+== Message format version
 
 When a producer sends a message to a Kafka broker, the message is encoded using a specific format. This is referred to as the message format version. You can configure a Kafka broker to convert messages from newer format versions to a given older format version, before the broker appends the message to the log.
 
@@ -40,7 +40,7 @@ In Kafka, there are two different methods for setting the message format version
 
 The default value of `message.format.version` for a  topic is defined by the `log.message.format.version` that is set on the Kafka broker. You can manually set the `message.format.version` of a topic by modifying its topic configuration. This chapter refers to the _message format version_ throughout when discussing both Kafka brokers and topics.
 
-Procedure outline::
+== Procedure outline
 
 In summary, upgrading to {ProductName} 1.1.0 is a three-stage process. To upgrade brokers and clients without downtime, you *must* complete the upgrade procedures in the order shown here.
 
--- a/books/assembly-upgrade-1-1-0.adoc
+++ b/books/assembly-upgrade-1-1-0.adoc
@@ -17,7 +17,7 @@ This chapter describes how to upgrade {ProductName} on Red Hat Enterprise Linux 
 |1.1.0                 |2.1.1
 |=======================
 
-Although {ProductName} 1.1.0 and Kafka 2.1.1 are both minor releases, the Kafka protocol has changed since Kafka 2.0.0 was released. In particular, the message format version and inter-broker protocol version are both now at version 2.1. As a result, the upgrade process involves making both configuration changes to existing Kafka brokers and code changes to client applications (consumers and producers). The following table shows the differences between the two Kafka versions:
+Although {ProductName} 1.1.0 and Kafka 2.1.1 are both minor releases, the Kafka protocol has changed since Kafka 2.0.0 was released. In particular, the message format version and inter-broker protocol version are both now at version 2.1. As a result, the upgrade process involves making configuration changes to existing Kafka brokers and code changes to client applications (producers, consumers, Kafka Connect, and MirrorMaker). The following table shows the differences between the two Kafka versions:
 
 [options="header"]
 |=======================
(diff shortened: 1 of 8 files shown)


laidan6000:
Should this refer to "Kafka Connect connectors" rather than "Kafka Connect"?

books/assembly-upgrade-1-1-0.adoc
```diff
@@ -17,7 +17,7 @@ This chapter describes how to upgrade {ProductName} on Red Hat Enterprise Linux
 |1.1.0                 |2.1.1
 |=======================
 
-Although {ProductName} 1.1.0 and Kafka 2.1.1 are both minor releases, the Kafka protocol has changed since Kafka 2.0.0 was released. In particular, the message format version and inter-broker protocol version are both now at version 2.1. As a result, the upgrade process involves making both configuration changes to existing Kafka brokers and code changes to client applications (consumers and producers). The following table shows the differences between the two Kafka versions:
+Although {ProductName} 1.1.0 and Kafka 2.1.1 are both minor releases, the Kafka protocol has changed since Kafka 2.0.0 was released. In particular, the message format version and inter-broker protocol version are both now at version 2.1. As a result, the upgrade process involves making configuration changes to existing Kafka brokers and code changes to client applications (producers, consumers, Kafka Connect, and MirrorMaker). The following table shows the differences between the two Kafka versions:
```
scholzj: Seems a bit inconsistent ... why _Kafka Connect_ but not _Kafka Mirror Maker_?

books/assembly-upgrade-1-1-0.adoc
```diff
@@ -49,9 +49,10 @@ In summary, upgrading to {ProductName} 1.1.0 is a three-stage process. To upgrad
 .. xref:proc-updating-zookeeper-binaries-{context}[Updating the Zookeeper binaries]
 
 . Second, upgrade all Kafka brokers to {ProductName} 1.1.0 and configure them to use the previous protocol versions.
+
 .. xref:proc-upgrading-kafka-brokers-to-amq-streams-1-1-0-{context}[Upgrading Kafka brokers to {ProductName} 1.1.0]
 
-. Third, upgrade all Kafka brokers and client applications to Kafka 2.1.1. To avoid cluster downtime, this stage involves:
+. Third, upgrade all Kafka brokers and client applications (producers, consumers, Kafka Connect, and MirrorMaker) to Kafka 2.1.1. To avoid cluster downtime, this stage involves:
```
scholzj: Same as above.

books/con-strategies-for-upgrading-clients.adoc
```diff
@@ -6,7 +6,7 @@
 
 = Strategies for upgrading clients
 
-The best approach to upgrading your client applications (including Kafka Connect connectors) depends on your particular circumstances.
+The best approach to upgrading your client applications (producers, consumers, Kafka Connect, and MirrorMaker) depends on your particular circumstances.
```
scholzj: Same as above.

scholzj:
@laidan6000 Well, it depends what do you want to update. In your text IMHO Kafka Connect is correct. 

However, from the title of this PR, I expected that this will have the upgrade guide for Kafka Connect and Kafka Mirror Maker. Which it doesn't. Connect and Mirror maker should really have their own upgrade procedures.

books/assembly-upgrade-1-1-0.adoc
PaulRMellor: My only question would be whether we need to repeat the same information in brackets in three places. Maybe we could say it once as something like `Client applications might include...` in Strategies for Upgrading Clients. If we need separate upgrade procedures for Connect and Mirror Maker, as Jakub suggests, then we'll need to reorganise this assembly. (Note, I'm working on the same content for PR #1528)

books/assembly-upgrade-1-1-0.adoc
```diff
@@ -28,7 +28,7 @@ Although {ProductName} 1.1.0 and Kafka 2.1.1 are both minor releases, the Kafka
 
 Although Kafka 2.0.0 and 2.1.1 use the same version of Zookeeper, Red Hat recommends that you update your Zookeeper cluster to use the newest Zookeeper binaries before proceeding with the main {ProductName} 1.1.0 upgrade.
```
tombentley:
```suggestion
Although Kafka 2.0.0 and 2.1.1 use the same version of Zookeeper, it is recommended that you update your Zookeeper cluster to use the newest Zookeeper binaries before proceeding with the main {ProductName} 1.1.0 upgrade.
```

books/proc-updating-kafka-brokers-to-new-message-format-version.adoc
```diff
@@ -6,7 +6,7 @@
 
 = Configuring Kafka brokers to use the new message format version
 
-When client applications have been upgraded, you can update the Kafka brokers to use the new message format version (2.1).
+When client applications (producers, consumers, Kafka Connect, and MirrorMaker) have been upgraded, you can update the Kafka brokers to use the new message format version (2.1).
```
tombentley: Should we use an attribute for this "new message format version", because it's likely to change again in the future and we'll likely forget to update to update all the places it has been used.

laidan6000:
@scholzj - Regarding having separate procedures, @tombentley is of the view that Kafka Connect and Mirror Maker do not need separate procedures. We should just highlight that they are types of clients.

tombentley:
> @tombentley is of the view that Kafka Connect and Mirror Maker do not need separate procedures. We should just highlight that they are types of clients.

It's different on OCP, of course, because we're providing the operators for these, so they warrant procedures imho, but on RHEL I don't think it's really necessary.

scholzj:
TBH, I do not agree with that. i think we should have a procedure for that on both. Even if it would be just a stupid procedure which would say to change the `spec.version` property on OCP or unzip the new binaries on RHEL. It would help to remind the users that they should upgrade it as well. 

For Mirror Maker it might also tell them how to configure different versions on each side. For Connect it should probably give some advice on what to do with connectors.

books/assembly-upgrade-1-1-0.adoc
laidan6000: I've gone with Paul's suggestion. Changed to Kafka MirrorMaker.

books/proc-updating-kafka-brokers-to-new-message-format-version.adoc
laidan6000: I suppose we'd also need an attribute for the inter-broker protocol versions. Given time constraints, I'd rather handle this when we make any necessary changes to this chapter to align with the next downstream release.

laidan6000:
Following Jakub's review, I've added a new procedure named "Upgrading Kafka Connect to AMQ Streams 1.1.0". This is mainly based on the existing procedure, "Upgrading Kafka brokers to AMQ Streams 1.1.0".

Decided not to have a standalone procedure for MirrorMaker upgrades because we are lacking in any other documentation on MirrorMaker.

Please review the changes on my file share: http://file.rdu.redhat.com/~dlaing/#assembly-upgrade-1-1-0-str

books/proc-upgrading-kafka-connect-to-amq-streams-1-1-0.adoc
```diff
@@ -0,0 +1,94 @@
...
+. Verify that Kafka Connect is running:
+
+** In standalone mode:
++
+[source,shell,subs=+quotes]
+----
+jcmd | grep ConnectStandalone
+----
+
+** In distributed mode
```
PaulRMellor:
```suggestion
** In distributed mode:
```

PR #56, opened 2019-08-28
Kafka authorization using SimpleACLAuthorizer

PaulRMellor:
New **_Authorization_** section for Configuring Kafka chapter.
We had some content that had not been added. I used that as a basis, reviewing against online material.

There is maybe more we want to say, but we have the basic configuration of SimpleACLAuthorizer.

### DIFF
--- a/books/assembly-configuring-kafka.adoc
+++ b/books/assembly-configuring-kafka.adoc
@@ -20,9 +20,9 @@ include::con-kafka-broker-id-configuration.adoc[leveloffset=+1]
 
 include::proc-running-multinode-kafka-cluster.adoc[leveloffset=+1]
 
-include::assembly-kafka-authorization.adoc[leveloffset=+1]
-
 include::assembly-kafka-zookeeper-authentication.adoc[leveloffset=+1]
+
+include::assembly-kafka-authorization.adoc[leveloffset=+1]
 
 include::assembly-kafka-zookeeper-authorization.adoc[leveloffset=+1]
 
--- a/books/assembly-configuring-kafka.adoc
+++ b/books/assembly-configuring-kafka.adoc
@@ -20,6 +20,8 @@ include::con-kafka-broker-id-configuration.adoc[leveloffset=+1]
 
 include::proc-running-multinode-kafka-cluster.adoc[leveloffset=+1]
 
+include::assembly-kafka-authorization.adoc[leveloffset=+1]
+
 include::assembly-kafka-zookeeper-authentication.adoc[leveloffset=+1]
 
 include::assembly-kafka-zookeeper-authorization.adoc[leveloffset=+1]
--- a/books/assembly-kafka-zookeeper-authorization.adoc
+++ b/books/assembly-kafka-zookeeper-authorization.adoc
@@ -6,7 +6,7 @@
 
 = Zookeeper authorization
 
-When authentication is enabled between Kafka and Zookeeper, you can use Access Control List (ACL) rules to automatically control access to Kafka records.
+When authentication is enabled between Kafka and Zookeeper, you can use Zookeeper Access Control List (ACL) rules to automatically control access to Kafka's metadata stored in Zookeeper.
 
 include::con-kafka-zookeeper-authorization-enabling-acls.adoc[leveloffset=+1]
 
--- a/books/assembly-kafka-zookeeper-authorization.adoc
+++ b/books/assembly-kafka-zookeeper-authorization.adoc
@@ -6,8 +6,7 @@
 
 = Zookeeper authorization
 
-When authentication is enabled between Kafka and Zookeeper, Kafka can be configured to automatically protect all its records with Access Control List (ACL) rules which will allow only the Kafka user to change the data.
-All other users will have read-only access.
+When authentication is enabled between Kafka and Zookeeper, you can use Access Control List (ACL) rules to automatically control access to Kafka records.
 
 include::con-kafka-zookeeper-authorization-enabling-acls.adoc[leveloffset=+1]
 
--- a/books/con-kafka-authorization-simple.adoc
+++ b/books/con-kafka-authorization-simple.adoc
@@ -78,8 +78,7 @@ You can apply Kafka ACLs to three types of resource:
 
 - Topics
 - Consumer groups
-- Cluster
-
+- Clusters
 
 .Supported operations
 
--- a/books/con-kafka-authorization-simple.adoc
+++ b/books/con-kafka-authorization-simple.adoc
@@ -6,63 +6,86 @@
 
 = Simple ACL authorizer
 
-Kafka has built-in authorizer plugin called `SimpleAclAuthorizer`.
-This plugin uses Zookeeper to store the ACL rules.
-To enable Authorization or ACLs, the property `authorizer.class.name` has to be specified.
-It has to contain a fully qualified name of the authorizer class.
-For the built-in authorizer, the fully qualified name is `kafka.security.auth.SimpleAclAuthorizer`:
+Authorizer plugins, including `SimpleAclAuthorizer`, are enabled through the `authorizer.class.name` property:
 
-[source]
+[source, properties]
+----
 authorizer.class.name=kafka.security.auth.SimpleAclAuthorizer
+----
+
+A fully-qualified name is required for the chosen authorizer.
+For `SimpleAclAuthorizer`, the fully-qualified name is `kafka.security.auth.SimpleAclAuthorizer`.
 
 .ACL rules
 
-The structure of ACL rules is: _Principal `P` is allowed / denied operation `O` on resource `R` from host
-`H`._ 
-When no rules are present for a given resource, all actions are denied.
-This behavior can be changed by setting the property `allow.everyone.if.no.acl.found` to `true` in the Kafka configuration file.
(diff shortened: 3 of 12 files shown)


books/assembly-kafka-zookeeper-authorization.adoc
```diff
@@ -6,8 +6,7 @@
 
 = Zookeeper authorization
 
-When authentication is enabled between Kafka and Zookeeper, Kafka can be configured to automatically protect all its records with Access Control List (ACL) rules which will allow only the Kafka user to change the data.
-All other users will have read-only access.
+When authentication is enabled between Kafka and Zookeeper, you can use Access Control List (ACL) rules to automatically control access to Kafka records.
```
tombentley: This is confusingly worded. "Kafka records" makes it sound like this authz is about protected messages in Kafka, but it's not. It's about protecting Kafka metadata that's stored in ZK using ZK acls.
```suggestion
When authentication is enabled between Kafka and Zookeeper, you can use Zookeeper Access Control List (ACL) rules to automatically control access to Kafka's metadata stored in Zookeeper.
```

books/con-kafka-authorization-simple.adoc
```diff
@@ -6,63 +6,86 @@
...
+----
+
+For example, a rule might be set so that user:
+
+[source, subs="+quotes"]
+----
+_John can view the topic `comments` from host 127.0.0.1_
+----
+
+Host is the IP address of the Kafka broker.
```
tombentley: No, it's the IP address of the machine that "John" is connecting _from_, which may happen to be the IP of a broker, but could also be a non-broker machine. Using John as the example is a little misleading, since often the principle will represent a producer or consumer application, rather than a human operator. It might be worth giving two examples to illustrate that principal covers both.

books/con-kafka-authorization-simple.adoc
```diff
@@ -6,63 +6,86 @@
...
+.Principals
+
+A _principal_ represents the identity of a user.
+The format of the ID depends on the authentication mechanism used by clients to connect to Kafka:
+
+* `User:ANONYMOUS` when connected without authentication.
+* `User:<username>` when connected using simple authentication mechanisms, such as PLAIN or SCRAM.
++
+For example `User:admin` or `User:user1`.
+* `User:<DN>` when connected using TLS client authentication.
```
tombentley: Perhaps DN is be a bit opaque here. Should we use `<DistinguishedName>` ?

books/con-kafka-authorization-simple.adoc
```diff
@@ -6,63 +6,86 @@
...
++
+For example `User:CN=user1,O=MyCompany,L=Prague,C=CZ`.
+* `User:<Kerberos username>` when connected using Kerberos.
+
+The _DN_ is the distinguished name from the client certificate.
```
tombentley: Oh you mention it here, but it would be better immediately after the use of DN above.

books/con-kafka-authorization-simple.adoc
```diff
@@ -6,63 +6,86 @@
...
-* Clients authenticated using Kerberos by default will have a Kafka username that is the primary part of their Kerberos principal.
-The `sasl.kerberos.principal.to.local.rules` property can be used to configure how the Kafka principal should be built from the Kerberos principal.
-
-Kafka ACLs can be applied to 3 different types of resource:
+You can apply Kafka ACLs to three types of resource:
 
 - Topics
-- Consumer Groups
+- Consumer groups
 - Cluster
```
tombentley:
```suggestion
- The cluster
```

?

books/con-kafka-zookeeper-authorization-enabling-acls.adoc
```diff
@@ -6,19 +6,29 @@
 
 = ACL Configuration
 
-Enforcement of ACL rules is controlled by the `zookeeper.set.acl` property in the `config/server.properties` Kafka configuration file and is disabled by default.
-To enabled the ACL protection set `zookeeper.set.acl` to `true`:
+Enforcement of ACL rules is controlled by the `zookeeper.set.acl` property in the `config/server.properties` Kafka configuration file.
```
tombentley:
```suggestion
Enforcement of Zookeeper ACL rules is controlled by the `zookeeper.set.acl` property in the `config/server.properties` Kafka configuration file.
```

PaulRMellor:
The list of supported operations differs from the one we have in OCP doc:

* Read
* Write
* Delete
* Alter
* Describe
* All
* IdempotentWrite
* ClusterAction
* Create
* AlterConfigs
* DescribeConfigs

books/assembly-configuring-kafka.adoc
```diff
@@ -20,6 +20,8 @@ include::con-kafka-broker-id-configuration.adoc[leveloffset=+1]
 
 include::proc-running-multinode-kafka-cluster.adoc[leveloffset=+1]
 
+include::assembly-kafka-authorization.adoc[leveloffset=+1]
```
scholzj: Logically, I would probably put this behind the chapter about authentication. Because without authentication everyone is User:ANONYMOUS and authorization is useless.

books/con-kafka-authorization-simple.adoc
scholzj: Thi is the IP address of the client, not the broker.

books/con-kafka-authorization-simple.adoc
```diff
@@ -6,63 +6,86 @@
...
-For example `User:CN=user1,O=MyCompany,L=Prague,C=CZ`.
-* Clients authenticated using Kerberos by default will have a Kafka username that is the primary part of their Kerberos principal.
-The `sasl.kerberos.principal.to.local.rules` property can be used to configure how the Kafka principal should be built from the Kerberos principal.
-
-Kafka ACLs can be applied to 3 different types of resource:
+You can apply Kafka ACLs to three types of resource:
 
 - Topics
-- Consumer Groups
+- Consumer groups
```
scholzj: I think there should be some more. At least the Transactional ID. Here is the full list of resources: http://kafka.apache.org/documentation/#resources_in_kafka

books/con-kafka-authorization-simple.adoc
```diff
@@ -93,98 +116,107 @@ Fields with `X` mark the supported operations.
...
+.ACL management options
+
+ACL rules are managed using the `bin/kafka-acls.sh` utility, which is provided as part of the Kafka distribution package.
+
+Use `kafka-acls.sh` parameter options to add, list and remove ACL rules, and perform other functions.
 
 |===
 |Option |Type | Description |Default
 
-|`--add`
```
scholzj: All these options actually need to be used with the `--`. So why remove it?

books/con-kafka-authorization-simple.adoc
```diff
@@ -93,98 +116,107 @@ Fields with `X` mark the supported operations.
...
 |===
 
-Management of ACL rules is done using `kafka-acls.sh` utility which is part of the Kafka distribution package.
-It can add, list and remove ACL rules.
-`kafka-acls.sh` contains three primary options for these functions:
+.ACL management options
+
+ACL rules are managed using the `bin/kafka-acls.sh` utility, which is provided as part of the Kafka distribution package.
+
+Use `kafka-acls.sh` parameter options to add, list and remove ACL rules, and perform other functions.
```
scholzj: We should probably update this table with the information from http://kafka.apache.org/documentation/#security_authz_cli

books/con-kafka-authorization-simple.adoc
PaulRMellor: Host is the IP address of the machine that John is connecting from.

In most cases, the user is a producer or consumer application:

[source, subs="+quotes"]
----
_Consumer01 can write to the consumer group `accounts` from host 127.0.0.1_
----

PaulRMellor: Fixed

PaulRMellor: Fixed

books/con-kafka-authorization-simple.adoc
PaulRMellor: I've added TransactionId and DelegationToken. If we prefer not to emphasize uncommon ones, we could introduce by saying:
'Kafka ACLs are commonly applied to these types of resource: ... ' 
'Other resources supported by Kafka include TransactionId and DelegationToken. '

books/con-kafka-authorization-simple.adoc
PaulRMellor: I thought it made the table options less easy to read. I added them back and I still think that's the case. I added this: The parameters require a double-hyphen convention, such as `--add`.

tombentley: I think it's more confusing without the `--`. The reader is interested to know the _option_ and the option is `--add`, not `add`.

books/con-kafka-authorization-simple.adoc
PaulRMellor: Updated. I guess there's value to having this information in the doc, but an alternative would be to link.

scholzj: If that works for the Red Hat Docs rules it is fine for me to link instead of rewriting this table.

PaulRMellor:
@scholzj - I am going to merge this content and request a QE review of the procedure.

PR #57, opened 2019-09-06
ACL authorization

PaulRMellor:
Following QE test of the procedures, clarification of the requirements for authentication before you define ACLs.

### DIFF
--- a/books/con-kafka-authorization-simple.adoc
+++ b/books/con-kafka-authorization-simple.adoc
@@ -69,7 +69,7 @@ You can use the `sasl.kerberos.principal.to.local.rules` property to configure h
 
 == Authentication of users
 
-If you wish to use authenticated users in your rules (not `User.ANONYMOUS` principals), set your authentication before you define your authorization rules.
+If you wish to use authenticated users in your rules (not `User:ANONYMOUS` principals), set your authentication before you define your authorization rules.
 
 For more information on methods of authentication, see xref:assembly-kafka-encryption-and-authentication-{context}[Encryption and authentication].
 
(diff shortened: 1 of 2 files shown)


books/con-kafka-authorization-simple.adoc
```diff
@@ -67,6 +67,12 @@ The _DistinguishedName_ is the distinguished name from the client certificate.
 The _Kerberos username_ is the primary part of the Kerberos principal, which is used by default when connecting using Kerberos.
 You can use the `sasl.kerberos.principal.to.local.rules` property to configure how the Kafka principal is built from the Kerberos principal.
 
+== Authentication of users
+
+If you wish to use authenticated users in your rules (not `User.ANONYMOUS` principals), set your authentication before you define your authorization rules.
```
scholzj:
```suggestion
If you wish to use authenticated users in your rules (not `User:ANONYMOUS` principals), set your authentication before you define your authorization rules.
```

scholzj: TBH this sentence looks a bit weird. What about something like this:

```suggestion
To use authorization, you need to have authentication enabled and used by your clients.
Otherwise all connections will have the principal `User:ANONYMOUS`.
```

books/con-kafka-authorization-simple.adoc
```diff
@@ -78,7 +84,13 @@ For example:
 [source]
 super.users=User:admin,User:operator
 
-.Supported resources
+.Replica broker authentication
```
scholzj: I would probably replace this while section with something like this:

```
.Replication between brokers

When authorization is enabled, it is applied to all listeners and all connections.
This includes the inter-broker connections used for replication of data between brokers.
When enabling authorization, you need to ensure that you use authentication for inter-broker connections and give the users used by the brokers sufficient rights.
For example, if authentication between brokers uses the `kafka-broker` user, then super user configuration must include the username `super.users=User:kafka-broker`.
```

PR #65, opened 2019-11-04
Make ZooKeeper references consistent

PaulRMellor:
ZooKeeper naming convention is now consistent in text.
Attribute `:Zookeeper: ZooKeeper` was created and used in the text to replace references to 'Zookeeper' and 'ZooKeeper'.

- Properties were kept as they are -- lowercase `zookeeper`
- Replaceables in code were updated/fixed
- Code formats were updated where the new attribute was introduced


PaulRMellor:
@scholzj Attribute replaced with text

PR #67, opened 2019-11-11
New architecture graphic and updated intro

PaulRMellor:
The architecture graphic in the Overview has been updated.
Now include Kafka Bridge and MirrorMaker
Intro text updated accordingly


ppatierno:
@PaulRMellor this PR contains all the stuff related to ZooKeepe renaming. Can you fix it before our review please? ;-) There are 52 files changed :-(

laidan6000:
Looks good.

books/assembly-overview.adoc
```diff
@@ -10,20 +10,24 @@
...
-It consists of the following main components:
 
-ZooKeeper:: Service for highly reliable distributed coordination.
-Kafka Broker:: Messaging broker responsible for delivering records from producing clients to consuming clients.
-Kafka Connect:: A toolkit for streaming data between Kafka brokers and other systems using _Connector_ plugins.
-Kafka Consumer and Producer APIs:: Java based APIs for producing and consuming messages to and from Kafka brokers.
+The main components comprise:
+
 Kafka Streams API:: API for writing _stream processor_ applications.
+Kafka Bridge:: A RESTful interface to AMQ Streams.
```
scholzj: I think it would be great to have some order to these - at least here in test it should be easy to fix:
* The Streams API and Consumer / Producer APIs are cliets and should be next to each other.  
* Zookeeper and Broker are the core compoenents and should be probably up on top. 
* Given how the Zookeeper is only an implementation detail from our perspective maybe it should be just mentioned as a comment next to the broker?
* Connect and Mirror Maker are official Kafka project components where as Bridge is our own - I feel like we should maybe distinguish it?
* Do we need to add the KAfka Exporter?

PaulRMellor: Thanks Jakub. Good suggestions. I will update the text and make a request for an updated graphic.

PR #70, opened 2019-12-16
Updates to oauth config chapter

PaulRMellor:
Signed-off-by: prmellor <pmellor@redhat.com>

_OAUTH 2.0_ 
Document use of additional OAuth2 configuration options.
Also include updates to the concepts to make the configuration options clearer.


PaulRMellor:
Reference doc: http://file.emea.redhat.com/pmellor/docs/RHEL/1-4-oauth-rhel.html#assembly-oauth-str

documentation/rhel/modules/con-kafka-authentication.adoc
```diff
@@ -6,9 +6,12 @@
...
 = Authentication
 
-Kafka supports two methods of authentication.
-On all connections, authentication using one of the supported SASL (Simple Authentication and Security Layer) mechanisms can be used.
-On encrypted connections, TLS client authentication based on X.509 certificates can be used.
+For authentication, you can use:
+
+* TLS client authentication based on X.509 certificates on encrypted connections
+* A supported SASL (Simple Authentication and Security Layer) mechanism on all connections
+* xref:assembly-oauth-str[{oauth} token based authentication]
```
scholzj: Technically, even this is a _supported SASL mechanism_. I'm not sure if we need to make it more clear or not.

documentation/rhel/modules/oauth/proc-oauth-broker-config.adoc
```diff
@@ -35,67 +38,69 @@ bcprov-*.jar
...
+. Configure the Kafka broker listener configuration in the `server.properties` file.
 +
 [source,xml,subs="+attributes,quotes"]
 ----
 sasl.enabled.mechanisms=OAUTHBEARER <1>
 
 listeners=CLIENTS://kafka:9092 <2>
 
-listener.security.protocol.map=CLIENTS:SASL_PLAINTEXT <3>
+listener.security.protocol.map=CLIENTS:SASL_SSL <3>
```
scholzj: The configuration snippet doesn't suggest that TLS is enabled here. So IMHO `SASL_PLAINTEXT` was better here.

documentation/rhel/using/build/en-US/master.xml
```diff
@@ -8670,7 +8670,7 @@ The port to publish to ZooKeeper for clients to use. In IaaS environments, this
 <emphasis role="strong">Default:</emphasis> -1<?asciidoc-br?>
 <emphasis role="strong">Importance:</emphasis> high<?asciidoc-br?>
 <emphasis role="strong">Dynamic update:</emphasis> read-only<?asciidoc-br?></simpara>
-<simpara>The broker id for this server. If unset, a unique broker id will be generated.To avoid conflicts between ZooKeeper generated broker id’s and user configured broker id’s, generated broker ids start from reserved.broker.max.id + 1.</simpara>
+<simpara>The broker id for this server. If unset, a unique broker id will be generated.To avoid conflicts between zookeeper generated broker id’s and user configured broker id’s, generated broker ids start from reserved.broker.max.id + 1.</simpara>
```
scholzj: Are these changes intended? They seem to mess up with the desired form of the ZooKeeper name as used accross the docs.

documentation/rhel/modules/oauth/proc-oauth-broker-config.adoc
```diff
@@ -35,67 +38,69 @@ bcprov-*.jar
...
 
 listener.name.clients.oauthbearer.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required; <4>
 listener.name.clients.oauthbearer.sasl.login.callback.handler.class=io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler
```
mstruk: Setting `listener.name.clients.oauthbearer.sasl.login.callback.handler.class` is only required if inter broker communication uses listener that is configured with SASL_OAUTHBEARER mechanism - meaning that broker are required to authenticate with OAuth2 when connecting to other brokers.

documentation/rhel/modules/con-kafka-authentication.adoc
PaulRMellor: I've added Marko's clarification. In terms of introducing a list of options, I think it's enough.

documentation/rhel/modules/oauth/proc-oauth-broker-config.adoc
PaulRMellor: @mstruk - Thanks for the clarification. Shall I assume no change required then?

documentation/rhel/using/build/en-US/master.xml
PaulRMellor: Pushed up by mistake.

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+.Authentication method as {oauth}
+[source,env,subs="+quotes, attributes"]
+----
+sasl.enabled.mechanisms=OAUTHBEARER
+----
+
+You then configure listeners with `plain` or `tls` xref:con-kafka-authentication-{context}[authentication],
+but it is recommended not to use `plain` listeners with {oauth} as this creates a vulnerability to network eavesdropping and unauthorized access through token theft.
+
```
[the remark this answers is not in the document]
PaulRMellor: I've removed this section and added the example configuration for TLS,

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,82 @@
...
+== {oauth} authentication configuration in the Kafka cluster
+
+To use {oauth} authentication in the Kafka cluster, you enable a TLS listener configuration for your Kafka cluster in the Kafka `server.properties` file:
+
+.Authentication method as {oauth}
+[source,env,subs="+quotes, attributes"]
+----
+sasl.enabled.mechanisms=OAUTHBEARER
+----
+
```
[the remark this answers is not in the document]
PaulRMellor: I've updated the concepts so that we now describe a minimum listener configuration for a broker and a TLS listener.

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+This means that inter-broker communication goes through the same listener as application clients.
+
+To enable {oauth} configuration for a specific listener, you specify `listener.name.clients.sasl.enabled.mechanisms` instead of `sasl.enabled.mechanisms`,
+which is shown in the example TLS listener configuration.
+
+.Minimum listener configuration for {oauth} authentication
+[source,env,subs="+quotes, attributes"]
+----
+sasl.enabled.mechanisms=OAUTHBEARER <1>
+listeners=CLIENT://kafka:9092 <2>
```
scholzj: This would probably listen on a hostname called `kafka`, which might not exist. The below is better option.

```suggestion
listeners=CLIENT://0.0.0.0:9092 <2>
```

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+listener.security.protocol.map=CLIENTS:SASL_PLAINTEXT <3>
```
scholzj:
```suggestion
listener.security.protocol.map=CLIENT:SASL_PLAINTEXT <3>
```

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+listener.name.client.sasl.enabled.mechanisms=OAUTHBEARER <4>
+sasl.mechanism.inter.broker.protocol=OAUTHBEARER <5>
```
scholzj: Does this actually really wotk for the replication? Doesn't this need some additional configuration to configure the client part of replication to use OAuth?

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+.TLS listener configuration for {oauth} authentication
+[source,env,subs="+quotes, attributes"]
+----
+sasl.enabled.mechanisms=
+listeners=REPLICATION://kafka:9091,CLIENT://kafka:9092 <1>
+listener.security.protocol.map=REPLICATION:SSL,CLIENT:SASL_PLAINTEXT <2>
+listener.name.client.sasl.enabled.mechanisms=OAUTHBEARER
+inter.broker.listener.name=REPLICATION
+ssl.keystore.password=_<my-keystore-password>_ <3>
```
scholzj: This could be a bit confusing since you configure password and type on the upper level but the location with the `listener.name.replication.` prefix. I think it should be configured both with the prefix to make it more clear to readers what does it relate to.

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+ssl.truststore.password=_<my-truststore-password>_
+ssl.keystore.type=PKCS12
+ssl.truststore.type=PKCS12
+ssl.endpoint.identification.algorithm=HTTPS
+ssl.secure.random.implementation=SHA1PRNG
+listener.name.replication.ssl.keystore.location=_<path-to-keystore-p12>_ <4>
+listener.name.replication.ssl.truststore.location=_<path-to-truststore-p12>_ <5>
+listener.name.replication.ssl.client.auth=required <6>
+listener.name.client.oauthbearer.sasl.server.callback.handler.class=io.strimzi.kafka.oauth.server.JaasServerOauthValidatorCallbackHandler
+listener.name.client.oauthbearer.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
```
scholzj: Single line again?

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+The endpoint contains the public keys used to validate signed JWT tokens, which are sent as credentials by Kafka clients.
+
+NOTE: All communication with the authorization server should be performed using HyperText Transfer Protocol Secure (HTTPS).
+
+For a TLS listener, you can configure a certificate _truststore_ and point to the truststore file.
+
+.Example environment variables for fast local JWT token validation
+[source,env,subs="+quotes, attributes"]
+----
+export OAUTH_VALID_ISSUER_URI=https://_<auth-server-address>_ <1>
```
scholzj: In the previous examples we configure everything through the JAAS configuration in the Kafka broker config file. Here we suddenly switch to environment variables. Would that be confusing for the user?

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+<6> The location of the truststore used in the TLS configuration.
+<7> Password to access the truststore.
+<8> The truststore type in PKCS #12 format.
+
+Instead of using environment variables, you can specify the configuration for the listener in the `server.properties` file.
+
+.Example properties file for fast local JWT token validation
+[source,env,subs="+quotes, attributes"]
+----
+listener.name.client.oauthbearer.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
```
scholzj: Single line again?

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+[source,env,subs="+quotes, attributes"]
+----
+sasl.enabled.mechanisms=
+listeners=REPLICATION://kafka:9091,CLIENT://kafka:9092 <1>
+listener.security.protocol.map=REPLICATION:SSL,CLIENT:SASL_PLAINTEXT <2>
+listener.name.client.sasl.enabled.mechanisms=OAUTHBEARER
+inter.broker.listener.name=REPLICATION
+ssl.keystore.password=_<my-keystore-password>_ <3>
+ssl.truststore.password=_<my-truststore-password>_
+ssl.keystore.type=PKCS12
```
scholzj: Should we be using JKS as the key format accross the RHEL docs? That is what is Java native, so it is more logical to use.

documentation/rhel/modules/oauth/proc-oauth-broker-config.adoc
```diff
@@ -35,67 +43,67 @@ bcprov-*.jar
...
 ----
-sasl.enabled.mechanisms=OAUTHBEARER <1>
-
-listeners=CLIENTS://kafka:9092 <2>
-
-listener.security.protocol.map=CLIENTS:SASL_PLAINTEXT <3>
-
-listener.name.clients.oauthbearer.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required; <4>
+sasl.enabled.mechanisms=OAUTHBEARER
+listeners=CLIENT://kafka:9092
```
scholzj:
```suggestion
listeners=CLIENT://0.0.0.0:9092
```

documentation/rhel/modules/oauth/proc-oauth-broker-config.adoc
```diff
@@ -35,67 +43,67 @@ bcprov-*.jar
...
+listener.security.protocol.map=CLIENTS:SASL_PLAINTEXT
```
scholzj:
```suggestion
listener.security.protocol.map=CLIENT:SASL_PLAINTEXT
```

documentation/rhel/modules/oauth/proc-oauth-broker-config.adoc
```diff
@@ -35,67 +43,67 @@ bcprov-*.jar
...
+listener.name.client.sasl.enabled.mechanisms=OAUTHBEARER
+sasl.mechanism.inter.broker.protocol=OAUTHBEARER
+inter.broker.listener.name=CLIENT
+listener.name.client.oauthbearer.sasl.server.callback.handler.class=io.strimzi.kafka.oauth.server.JaasServerOauthValidatorCallbackHandler
+listener.name.client.oauthbearer.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
```
scholzj: Should this have `;` at the end instead of `\`?

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
scholzj: In most cases `0.0.0.0` would work fine and use the actual IP as advertised host. It has definitely much better chances on working than `kafka`.

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
scholzj: In Strimzi nobody expects the user to write his own configuration file. Here we do, so we need to make sure the examples are understandable. here is not much to test, we just need to prefix the rest of the TLS configurations as well.

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
scholzj: Ok, if this works than it is all fine with me of course. 👍

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+[source,env,subs="+quotes, attributes"]
+----
+sasl.enabled.mechanisms=OAUTHBEARER <1>
+listeners=CLIENT://kafka:9092 <2>
+listener.security.protocol.map=CLIENTS:SASL_PLAINTEXT <3>
+listener.name.client.sasl.enabled.mechanisms=OAUTHBEARER <4>
+sasl.mechanism.inter.broker.protocol=OAUTHBEARER <5>
+inter.broker.listener.name=CLIENT <6>
+listener.name.client.oauthbearer.sasl.server.callback.handler.class=io.strimzi.kafka.oauth.server.JaasServerOauthValidatorCallbackHandler <7>
+listener.name.client.oauthbearer.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \ <8>
```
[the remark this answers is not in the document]
scholzj: Ok, if this works than it is all fine with me of course. 👍

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
scholzj: I do not think we need to document it just because. If we do not think the users should not use it, we can easily not mention it. That would give us at least some freedom to change the names if needed. However, my main concern is that I didn't found any logic and any explanation here. It seems to just randomly change between one method and the other. So that might cause confusion.

PaulRMellor: Do we want to retain the environment variables examples?

scholzj: I can live with it as long as it is not confusing for the readers. I think the `con-oauth-authentication-broker.adoc` does it quite well to offering both options. My only caveat is that by documenting htem, they become sort of an API which we need to be careful to break. But I guess Marko has no plans to change them.

mstruk: They do, as you point out, jump out of nowhere in the doc. At least an intro paragraph is needed:

OAuth settings can be configured as JAAS module configuration, as we've seen previously, where we pass `key=value` pairs as parameters in a specially formated value of `listener.name.LISTENERNAME.oauthbearer.sasl.jaas.config` property.
Alternatively, we can specify the same `key=value` configuration as ENV variables. We can either use the same keys or allcaps ENV vars style.

In such a case we still have to specify  in 'server.properties' the JAAS configuration:

`listener.name.client.oauthbearer.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required ;`

We can omit all configuration parameter since they will be read from ENV vars.

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+<5> The user name profile claim (or key) that contains the actual user name in the token.
+The user name is the _principal_ used to identify the user.
+The value will depend on the authentication flow and the authorization server used.
+<6> The location of the truststore used in the TLS configuration.
+<7> Password to access the truststore.
+<8> The truststore type in PKCS #12 format.
+
+Instead of using environment variables, you can specify the configuration for the listener in the `server.properties` file.
+
+.Example properties file for fast local JWT token validation
```
mstruk: Maybe reorder, and move this chapter up, and then continue with ENV vars example intro paragraph.

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+<1> The {oauth} introspection endpoint URL. For example, _\https://<auth-server-address>/auth/realms/master/protocol/openid-connect/token/introspect_.
+<2> Client ID of the Kafka broker.
+<3> Secret for the Kafka broker.
+<4> The location of the truststore used in the TLS configuration.
+<5> Password to access the truststore.
+<6> The truststore type in PKCS #12 format.
+
+Instead of using environment variables, you can specify the configuration for a listener in the `server.properties` file.
+
+.Example properties file for an introspection endpoint
```
mstruk: Again, maybe reorder - putting it before ENV var example.

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
PaulRMellor: I updated the single property.

mstruk:
Is the idea for:

* https://github.com/PaulRMellor/strimzi-kafka-operator/blob/oauth-additional-options/documentation/assemblies/oauth/assembly-oauth.adoc
and
* https://github.com/PaulRMellor/amq-streams-docs/blob/oauth-additional-options/documentation/rhel/assemblies/oauth/assembly-oauth.adoc

To be the same?
Sounds pretty tedious to keep these in-sync manually because of tiny differences. And currently these two are out of sync.
I assume changes are made in upstream, and then copied to product docs. For parts that have to be in-sync it should suffice for review to be done on upstream, rather than repeating the review comments on product as well?

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
PaulRMellor: Thanks Marko. I've added a new section to the chapter.

image

documentation/rhel/modules/con-kafka-authentication.adoc
```diff
@@ -6,9 +6,12 @@
 
 = Authentication
 
-Kafka supports two methods of authentication.
-On all connections, authentication using one of the supported SASL (Simple Authentication and Security Layer) mechanisms can be used.
-On encrypted connections, TLS client authentication based on X.509 certificates can be used.
+For authentication, you can use:
+
+* TLS client authentication based on X.509 certificates on encrypted connections
+* A supported Kafka SASL (Simple Authentication and Security Layer) mechanism on all connections
```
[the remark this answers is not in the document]
PaulRMellor: Dropped. :+1:

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
+// Module included in the following assemblies:
+//
+// assembly-oauth.adoc
+
+[id='con-oauth-authentication-broker-{context}']
+= {oauth} Kafka broker configuration
+
+Kafka broker configuration for {oauth} involves:
+
+* Client configuration (for the Kafka broker) on an authorization server
```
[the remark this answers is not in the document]
PaulRMellor: The original reads fine to me. We don't actually say 'configuration involves configuration'. The bullets also match the following titles for the sub-sections.

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+* Client ID and secret as the authentication mechanism
+
+== {oauth} authentication configuration in the Kafka cluster
+
+To use {oauth} authentication in the Kafka cluster, you enable a listener configuration for your Kafka cluster in the Kafka `server.properties` file.
+A minimum configuration is required.
+You can also configure a TLS listener, where TLS is used for inter-broker communication.
+
+The minimum configuration shown here applies a _global_ listener configuration.
+This means that inter-broker communication goes through the same listener as application clients.
```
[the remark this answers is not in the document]
PaulRMellor: I think we can leave this the same, not to confuse the reader, but I've added to callout 5 as follows: `The `inter.broker.listener.name` specification is required for the configuration to be valid.`

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+  oauth.username.claim="preferred_username"  \
+  unsecuredLoginStringClaim_sub="thePrincipalName" ; <7>
+----
+<1> Separate configurations are required for inter-broker communication and client applications.
+<2> Separate channel protocols are required for inter-broker communication and client applications.
+<3> The `ssl.` properties define the TLS configuration.
+<4> Path to the keystore for the listener.
+<5> Path to the truststore for the listener.
+<6> Specifies that TLS is required for the listener.
+<7> Configures the listener so that TLS client authentication with certificates is used for inter-broker communication.
```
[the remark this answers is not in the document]
PaulRMellor: <7> Configures the _CLIENT_ listener for {oauth}. Connectivity with the authorization server should use secure HTTPS connections.

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+
+== Fast local JWT token validation configuration
+
+Fast local JWT token validation checks a JWT token signature locally.
+
+The local check ensures a token:
+
+* Conforms to type (`oauth`)
```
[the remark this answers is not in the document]
PaulRMellor: We caught it in the other doc. I've updated as follows: Conforms to type by containing a (_typ_) claim value of `Bearer` for an access token

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,201 @@
...
+* Is valid (not expired)
+* Has an issuer that matches a valid issuer URI
+
+You specify a _valid issuer URI_ when you configure the listener environment variables, so that any tokens not issued by the authorization server are rejected.
```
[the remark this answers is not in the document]
PaulRMellor: Changed to _You specify a _valid issuer URI_ when you configure the listener, ..._

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
PaulRMellor: Yes. I thought that from earlier comments and already made the change. :+1:

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
PaulRMellor: Properties now come first.

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+* Creating the {oauth} client in the authorization server
+* Configuring {oauth} authentication in the Kafka custom resource
+
+NOTE: In relation to the authorization server, Kafka brokers and Kafka clients are both regarded as {oauth} clients.
+
+== {oauth} client configuration on an authorization server
+
+To configure a Kafka broker to validate the token received during session initiation,
+the recommended approach is to create a {oauth} _client_ definition in an authorization server, configured as _confidential_, with client credentials enabled:
```
laidan6000:
```suggestion
the recommended approach is to create a {oauth} _client_ definition in an authorization server, configured as _confidential_, with the following client credentials enabled:
```

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,203 @@
...
+----
+<1> Enables the _OAUTHBEARER_ as _SASL_ mechanism for credentials exchange over SASL.
+<2> Configures a listener for client applications to connect. The system `hostname` is used as an advertised hostname, which clients must resolve in order to reconnect.
+<3> Specifies the channel protocol for the listener. `SASL_SSL` is for TLS. `SASL_PLAINTEXT` is used for an unencrypted connection (no TLS), but there is risk of eavesdropping and interception at the TCP connection layer.
+<4> Specifies _OAUTHBEARER_ as _SASL_ for client listeners.
+<5> Specifies _OAUTHBEARER_ as _SASL_ for inter-broker communication.
+<6> Specifies the listener for inter-broker communication. The specification is required for the configuration to be valid.
+<7> Configures {oauth} authentication on the client listener.
+<8> Configures authentication settings for client and inter-broker communication.
+The `oauth.client.id`, `oauth.client.secret` and `auth.token.endpoint.uri` properties relate to inter-broker configuration.
```
laidan6000:
```suggestion
The `oauth.client.id`, `oauth.client.secret`, and `auth.token.endpoint.uri` properties relate to inter-broker configuration.
```

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,203 @@
...
+<4> Path to the keystore for the listener.
+<5> Path to the truststore for the listener.
+<6> Specifies that clients of the _REPLICATION_ listener have to authenticate with a client certificate when establishing a TLS connection (used for inter-broker connectivity).
+<7> Configures the _CLIENT_ listener for {oauth}. Connectivity with the authorization server should use secure HTTPS connections.
+
+== Fast local JWT token validation configuration
+
+Fast local JWT token validation checks a JWT token signature locally.
+
+The local check ensures a token:
```
laidan6000:
```suggestion
The local check ensures that a token:
```

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,203 @@
...
+<5> The token claim (or key) that contains the actual user name in the token.
+The user name is the _principal_ used to identify the user.
+The value will depend on the authentication flow and the authorization server used.
+<6> The location of the truststore used in the TLS configuration.
+<7> Password to access the truststore.
+<8> The truststore type in PKCS #12 format.
+
+== {oauth} introspection endpoint configuration
+
+Token validation using {oauth} an introspection endpoint treats a received access token as opaque.
```
laidan6000: Possible typo:

"Token validation using {oauth} AS an introspection endpoint..."

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,203 @@
...
+The Kafka broker sends an access token to the introspection endpoint, which responds with the token information necessary for validation.
+Importantly, it returns up-to-date information if the specific access token is valid, and also information about when the token expires.
+
+To configure {oauth} introspection-based validation, you specify an _introspection endpoint URI_ rather than the JWKs endpoint URI specified for fast local JWT token validation.
+Depending on the authorization server, you typically have to specify a _client ID_ and _client secret_, as the introspection endpoint is usually protected.
```
laidan6000:
```suggestion
Depending on the authorization server, you typically have to specify a _client ID_ and _client secret_, because the introspection endpoint is usually protected.
```

documentation/rhel/modules/oauth/con-oauth-authentication-client.adoc
```diff
@@ -0,0 +1,25 @@
...
+
+* The Credentials required to obtain a valid access token from an authorization server (client ID and Secret)
+* A valid long-lived access token or refresh token, obtained using tools provided by an authorization server
+
+Credentials are never sent to the Kafka broker.
+The only information ever sent to the Kafka broker is an access token.
+When a client obtains an access token, no further communication with the authorization server is needed.
+
+The simplest mechanism is authentication with a client ID and Secret.
+Using a long-lived access token, or a long-lived refresh token, adds more complexity as there is additional dependency on authorization server tools.
```
laidan6000:
```suggestion
Using a long-lived access token, or a long-lived refresh token, adds more complexity because there is additional dependency on authorization server tools.
```

documentation/rhel/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -5,36 +5,17 @@
...
-
-NOTE: If you are using long-lived access tokens, you can set policy in the authorization server to increase the maximum lifetime of the token.
-
-If the Kafka client is not configured with an access token directly, the client exchanges credentials for an access token during Kafka session initiation by contacting the authorization server.
-The Kafka client uses one of two mechanisms:
-
-* Client id and secret
-* Client id, refresh token, and (optionally) a secret
+* JAAS properties are configured in the `server.properties` configuration file, and passed as key-values pairs of the `listener.name._<listener-name>_.oauthbearer.sasl.jaas.config` property.
+* Using environment variables, you still need the `listener.name._<listener-name>_.oauthbearer.sasl.jaas.config` in the `server.properties` file, but you can drop the properties.
```
laidan6000:
```suggestion
* Using environment variables, you still need the `listener.name._<listener-name>_.oauthbearer.sasl.jaas.config` in the `server.properties` file, but you can omit the properties.
```

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
PaulRMellor: Definite typo :grin:

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,203 @@
...
+//
+// assembly-oauth.adoc
+
+[id='con-oauth-authentication-broker-{context}']
+= {oauth} Kafka broker configuration
+
+Kafka broker configuration for {oauth} involves:
+
+* Creating the {oauth} client in the authorization server
+* Configuring {oauth} authentication in the Kafka custom resource
```
ppatierno: it's not about Kubernetes but just RHEL, so no "custom resource"

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+NOTE: In relation to the authorization server, Kafka brokers and Kafka clients are both regarded as {oauth} clients.
+
+== {oauth} client configuration on an authorization server
+
+To configure a Kafka broker to validate the token received during session initiation,
+the recommended approach is to create a {oauth} _client_ definition in an authorization server, configured as _confidential_, with the following client credentials enabled:
+
+* Client ID of `kafka`
```
ppatierno: here we are saying that the client id is `kafka` while in the next paragraph we have `oauth.client.id="kafka-broker"`.
I think it should be `kafka-broker` here as well.

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
```diff
@@ -0,0 +1,203 @@
...
+* Is valid (not expired)
+* Has an issuer that matches a `validIssuerURI`
+
+You specify a _valid issuer URI_ when you configure the listener, so that any tokens not issued by the authorization server are rejected.
+
+The authorization server does not need to be contacted during fast local JWT token validation.
+You activate fast local JWT token validation by specifying a _JWKs endpoint URI_ exposed by the {oauth} authorization server.
+The endpoint contains the public keys used to validate signed JWT tokens, which are sent as credentials by Kafka clients.
+
+NOTE: All communication with the authorization server should be performed using HyperText Transfer Protocol Secure (HTTPS).
```
ppatierno: to be honest I think that we can just use HTTPS here without the full name

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
PaulRMellor: _...Kafka cluster_

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
PaulRMellor: Yeah. I've also put _(for example)_

documentation/rhel/modules/oauth/con-oauth-authentication-broker.adoc
PaulRMellor: At this level, I guess so. :+1:

PR #71, opened 2021-07-28
Update for Kafka 2.8

tombentley:


PaulRMellor:
@tombentley I updated the pom downstream and rebuilt. A couple of config updates, but the MBeans were actually up-to-date, so no changes (checking against the changes here).

